PR #2, opened 2019-03-25
added pfs doc from hackmd

corpetty:
Pedro created this a long time ago... adding here to start this off.  Maybe we should edit it a bit to conform to some standard, but wanted to start something.

### DIFF
--- a/pfs.md
+++ b/pfs.md
@@ -1,0 +1,234 @@
+Status Perfect Forward Secrecy Whitepaper
+====
+
+[TOC]
+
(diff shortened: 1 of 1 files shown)


oskarth:
Thanks! Agree we should add this. Let's figure out format first so it is proper from the start. I imagine the first two to:

1) Capture process for draft/accepted etc
2) Capture current state, or at least partial state, to such an extent that anyone can implement Status

oskarth:
Closing as this is captured in existing doc.

PR #3, opened 2019-04-01
[WIP] add Status Secure And Decentralized Messaging Protocol draft

adambabik:
Current status: draft

This is a specification of the currently implemented messaging protocol in Status Mobile and Status Desktop.

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -1,0 +1,177 @@
+Status Secure And Decentralized Messaging Protocol
+==================================================
+
+# Abstract
+
+Ethereum empowers users and developers to interact with totally new kind of applications called Dapps (Decentralized applications). These application allows to interact with the blockchain on a completely new level which is not only about exchanging values but also executing arbitrary logic. This logic can form very sophisticated programs like DAOs (Decentralized autonomous organizations). The missing part here is how users of Dapps can communicate securely and in a decentralized way with each other. Communication is an essential part of any activity. In this document, we specify a secure and decentralized messaging protocol that is capable of running on the Ethereum network.
+
+# Introduction
+
+TBD
+
+# Terminology
+
+TBD
+
+# Basic Assumption
+
+This protocol assumes that there is an Ethereum node that is capable of discoverying peers and implements Whisper protocol. It also assumes that the participants of a given Whisper network accept messages with lowered PoW value.
+
+# Protocol Overview
+
+Notice: this protocol is documented post factum. The goal of it is to clearly present the current design and prepare the ground for its second version.
+
+The implementation of this protocol is mainly done in https://github.com/status-im/status-react and https://github.com/status-im/status-go repositories.
+
+The goal of this protocol is to allow people running Ethereum nodes with Whisper service enabled to exchange messages that are encrypted end-to-end in a way that guarantees darkness to some extent (messages are broadcasted through the network rather than send directly or using a known path).
+
+It's important to notice that messages are not limited to be text messages only. They can also have special meaning depending on the client implementation. For example, in the current implementation, there are message which informs about Eth requests or current user's location.
+
+This protocol consist of two components:
+* payload specification
+* Whisper adapter.
+
+The payload specification describes how the messages are encoded and decoded and what each fields mean. This is required to properly interpret messages by the client.
+
+Whisper adapter specifies interaction with the Whiper service with regards to keys management, configuration and some metadata (like topic) required to properly process and encrypt/decrypt messages.
+
+The protocol does not specify things like peers discovery, however, some notes and the current implementation will be described in an appendix.
+
+# Payload
+
+Payload is encrypted and decrypted using [transit format](https://github.com/cognitect/transit-format). It is a text-based format so that each decrypted message's payload can be easily read.
+
+Payload contains the following fields:
+* text `string`
+* content type `enum { text/plain }`
+* message type `enum { public-group-user-message, user-message }`
+* clock `int64`
+* timestamp `int64`
+* content `struct { chat-id string, text string }`
+
+Example of a valid message:
+
+```
+["~#c4",["abc123","text/plain","~:public-group-user-message",154593077368201,1545930773682,["^ ","~:chat-id","testing-adamb","~:text","abc123"]]]
+```
+
+As you can see, the message is an array and each index value has its meaning:
+* 0: `c4` is a decoder handler identificator for the current payload format. Identificators allow to register handler for many different types of messages
+* 1: array which items correspond to the described payload fields above
+
+For more details regarding serialization and deserialization please consult [transit format](https://github.com/cognitect/transit-format) specification.
+
+## Content types
+
+Here is a list of supported content types:
+* `text/plain` identifies a message which content is a plain text
+
+TODO: are there other types? How is it useful? Sending Eth requests?
+
+## Message types
+
+Here is a list of supported message types:
+* `public-group-user-message` is a message to the public group
+* `user-message` is a private message.
+
+TODO: how is it useful? What are other supported message types?
+
+## Message ordering
+
+TBD
+
+## Quoting
+
+TBD
+
+# Whisper adapter
+
+Whisper has been chosen as an messages exchange protocol because it was designed as an off-chain communcation layer for the Ethereum nodes. It supports e2e encryption and uses epidemic spread to route data to all members of the network. It also provides [darkness to some extent](https://github.com/ethereum/go-ethereum/wiki/Achieving-Darkness).
+
+However, Whisper was not designed to handle huge number of messages and real-time communication. These are the tradeoffs that we accepted when implementing this protocol.
+
+This protocol can either work using a Whisper service which requires a protocol implementation to run in the same process as a Whisper node as well as using a Whisper client which might run as a separate Ethereum node and communicate through IPC or WebSocket.
+
+There is some tight coupling between the payload and Whisper:
+* Whisper message topic depends on the actual message type
+* Whisper message uses a different key (asymmetric or symmetric) depending on the actual message type
+
+## Whisper node configuration
+
+If you want to run a Whisper node and receive messages from Status clients, you need to configure it properly, otherwise, some message might not be received.
+
+Whisper's Proof Of Work algorithm is used to to deter denial of service and various spam/flood attacks against the Whisper network. The sender of a message must perform some work which in this case means performing some computations. Because Status main client is a mobile client, this easily leads to battery draining and poor performance of the app itself. Hence, these clients send envelopes with the following settings:
+* proof-of-work equal `0.002`
+* proof-of-work time equal to `10`
+* time-to-live equal `10`
+
+If you want to receive messages from a mobile client, you need to run Whisper node with PoW set to `0.002`. In case of `geth`, this option can be overriden with `-shh.pow=0.002` flag.
+
+## Topic
+
+There are two types of Whisper topics the protocol uses:
+* static topic for private chats
+* dynamic topic based on a chat name for public chats
+
+Static topic is always the same and its hex representation is `0xf8946aac`. It is also called a discovery topic. Having only one topic for all private chats has an advantage as it's very hard to figure out who talks to who. A drawback is that everyone receives everyone's messages but they can decrypt only these they have private keys for.
+
+A dynamic topic is derived from a string using the following algorithm:
+
+```
+var hash []byte
+
+hash = keccak256(name)
+
+# Whisper V6 specific
+var topic [4]byte
+
+topic_len = 4
+
+if len(hash) < topic_len {
+    topic_len = len(hash)
+}
+
+for i = 0; i < topic_len; i++ {
+    topic[i] = hash[i]
+}
+```
+
+## Encryption
+
+Protocol distinguishes messages encrypted using asymmetric and symmetric encryptions.
+
+Symmetric keys are created using `shh_generateSymKeyFromPassword` Whisper JSON-RPC method which accepts one param, a string.
+
+Messages encrypted with asymmetric encryption should be encrypted using recipeint's public key so that only the recipient can decrypt them.
+
+## Offline messages
+
+TBD
+
+# Perfect Forward Secrecy (PFS)
+
+TBD
+
+# Device syncing
+
+TBD
+
+# One-to-one messages
+
+One-to-one messages are also known as private messages.
+
+# Public messages
+
+TBD
+
+# Group messages
+
+TBD
+
+# Offline messages
+
+TBD
+
+# Whisper V6 extensions
+
+TBD
--- a/protocol.md
+++ b/protocol.md
@@ -2,7 +2,7 @@
 eip: 1
 title: Status Secure Messaging Protocol
 status: Draft
-type: Standard
+type: Standard Core
 author: Adam Babik <adam@status.im>
 created: 2019-04-18
 updated: 2019-04-18
--- a/protocol.md
+++ b/protocol.md
@@ -1,3 +1,13 @@
+---
+eip: 1
+title: Status Secure Messaging Protocol
+status: Draft
+type: Standard
+author: Adam Babik <adam@status.im>
+created: 2019-04-18
+updated: 2019-04-18
+---
+
 Status Secure And Decentralized Messaging Protocol
 ==================================================
 
@@ -52,7 +62,7 @@ Ethereum empowers users and developers to interact with totally new kind of appl
 This protocol assumes the following:
 1. There MUST be an Ethereum node that is capable of discovering peers and implements Whisper V6 specification.
 2. Participants of a given Whisper network in order to communicate with each other MUST accept messages with lowered PoW value. More in (Whisper node configuration)(#whisper-node-configuration).
-3. Time MUST be synced between all nodes participating in the given network (this is intrinsic requirement of the Whisper specification).
+3. Time MUST be synced between all nodes participating in the given network (this is intrinsic requirement of the Whisper specification as well). A clock drift between two peers larger than 20 seconds MAY result in discarding incoming messages.
 
 # Protocol Overview
 
@@ -350,4 +360,4 @@ All described addons are implemented in [status-im/whisper fork](https://github.
 
 ## New RPC methods
 
-TODO: provide a list of RPC methods
+TODO: provide a list of RPC methods from `shhext` API which are relevant to this spec.
--- a/protocol.md
+++ b/protocol.md
@@ -1,6 +1,30 @@
 Status Secure And Decentralized Messaging Protocol
 ==================================================
 
+- [Status Secure And Decentralized Messaging Protocol](#status-secure-and-decentralized-messaging-protocol)
+- [Abstract](#abstract)
+- [Introduction](#introduction)
+- [Terminology](#terminology)
+- [Basic Assumption](#basic-assumption)
+- [Protocol Overview](#protocol-overview)
+- [Payload](#payload)
+	- [Content types](#content-types)
+	- [Message types](#message-types)
+	- [Message ordering](#message-ordering)
+	- [Quoting](#quoting)
+- [Whisper adapter](#whisper-adapter)
+	- [Whisper node configuration](#whisper-node-configuration)
+	- [Keys management](#keys-management)
+	- [Topic](#topic)
+	- [Encryption](#encryption)
+- [Perfect Forward Secrecy (PFS)](#perfect-forward-secrecy-pfs)
+- [Device syncing](#device-syncing)
+- [One-to-one messages](#one-to-one-messages)
+- [Public messages](#public-messages)
+- [Group messages](#group-messages)
+- [Offline messages](#offline-messages)
+- [Whisper V6 extensions](#whisper-v6-extensions)
+
 # Abstract
 
 Ethereum empowers users and developers to interact with totally new kind of applications called Dapps (Decentralized applications). These application allows to interact with the blockchain on a completely new level which is not only about exchanging values but also executing arbitrary logic. This logic can form very sophisticated programs like DAOs (Decentralized autonomous organizations). The missing part here is how users of Dapps can communicate securely and in a decentralized way with each other. Communication is an essential part of any activity. In this document, we specify a secure and decentralized messaging protocol that is capable of running on the Ethereum network.
--- a/protocol.md
+++ b/protocol.md
@@ -8,10 +8,10 @@ created: 2019-04-18
 updated: 2019-04-18
 ---
 
-Status Secure And Decentralized Messaging Protocol
+Status Secure Messaging Protocol
 ==================================================
 
-- [Status Secure And Decentralized Messaging Protocol](#status-secure-and-decentralized-messaging-protocol)
+- [Status Secure Messaging Protocol](#status-secure-messaging-protocol)
 - [Abstract](#abstract)
 - [Terminology](#terminology)
 - [Basic Assumption](#basic-assumption)
--- a/protocol.md
+++ b/protocol.md
@@ -48,7 +48,7 @@ TBD
 * *Status Whisper node*: an Ethereum node with Whisper V6 enabled and additional Whisper extensions described in [Whisper V6 extensions (or Status Whisper Node)](#whisper-v6-extensions-or-status-whisper-node)
 * *MailServer*: an Ethereum node with Whisper V6 enabled and a mail server registered capable of storing and providing offline messages
 * *Message*: decrypted Whisper message
-* *Envelope*: encrypted message with some metadata like topic and TTL echanged between Whisper nodes; a symmetric or assymetric key is needed to decrypt it and read the payload
+* *Envelope*: encrypted message with some metadata like topic and TTL sent between Whisper nodes; a symmetric or asymmetric key is needed to decrypt it and read the payload
 * *Offline message*: an expired envelope stored by a Whisper node permanently
 
 # Basic Assumption
@@ -56,7 +56,7 @@ TBD
 This protocol assumes the following:
 1. There is an Ethereum node that is capable of discovering peers and implements Whisper V6 service.
 2. Participants of a given Whisper network in order to communicate with each other accept messages with lowered PoW value.
-3. Time is synced between all nodes participaiting in the given network (this is intrinsic Whisper requirement).
+3. Time is synced between all nodes participating in the given network (this is intrinsic Whisper requirement).
 
 # Protocol Overview
 
@@ -75,7 +75,7 @@ This protocol consist of three components:
 
 The payload specification describes how the messages are encoded and decoded and what each fields mean. This is required to properly interpret messages by the client.
 
-Whisper adapter specifies interaction with the Whiper service with regards to keys management, configuration and some metadata (like topic) required to properly process and encrypt/decrypt messages.
+Whisper adapter specifies interaction with the Whisper service with regards to keys management, configuration and some metadata (like topic) required to properly process and encrypt/decrypt messages.
 
 Offline messaging describes how the protocol handles delivering messages when one or more participants were offline and the messages expired in the network.
 
@@ -100,7 +100,7 @@ Example of a valid message:
 ```
 
 As you can see, the message is an array and each index value has its meaning:
-* 0: `c4` is a decoder handler identificator for the current payload format. Identificators allow to register handler for many different types of messages
+* 0: `c4` is a decoder handler identification for the current payload format. Identifications allow to register handler for many different types of messages
 * 1: array which items correspond to the described payload fields above
 
 For more details regarding serialization and deserialization please consult [transit format](https://github.com/cognitect/transit-format) specification.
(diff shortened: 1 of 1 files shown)


oskarth:
@adambabik Please ping / request review whenever you want a review on this

protocol.md
```diff
@@ -0,0 +1,196 @@
...
+TBD
+
+# Terminology
+
+* *Client*: a Whisper node implementing the protocol
+* *Whisper node*: an Ethereum node with Whisper V6 enabled (in the case of geth, it's `--shh` option)
+* *MailServer*: an Ethereum node with Whisper V6 enabled and a mail server registered capable of storing and providing offline messages
+* *Message*: decrypted Whisper message
+* *Envelope*: encrypted message with some metadata like topic and TTL echanged between Whisper nodes; a symmetric or assymetric key is needed to decrypt it and read the payload
+* *Offline message*: an expired envelope stored by a Whisper node permanently
```
mandrigin, line 58: I would also add info about a Fleet there.

protocol.md
```diff
@@ -0,0 +1,196 @@
...
+Ethereum empowers users and developers to interact with totally new kind of applications called Dapps (Decentralized applications). These application allows to interact with the blockchain on a completely new level which is not only about exchanging values but also executing arbitrary logic. This logic can form very sophisticated programs like DAOs (Decentralized autonomous organizations). The missing part here is how users of Dapps can communicate securely and in a decentralized way with each other. Communication is an essential part of any activity. In this document, we specify a secure and decentralized messaging protocol that is capable of running on the Ethereum network.
+
+# Introduction
+
+TBD
+
+# Terminology
+
+* *Client*: a Whisper node implementing the protocol
+* *Whisper node*: an Ethereum node with Whisper V6 enabled (in the case of geth, it's `--shh` option)
```
mandrigin, line 52: Maybe it is also worth mentioning the difference between Status Whisper nodes and regular whisper nodes.

protocol.md
```diff
@@ -0,0 +1,196 @@
...
+* *MailServer*: an Ethereum node with Whisper V6 enabled and a mail server registered capable of storing and providing offline messages
+* *Message*: decrypted Whisper message
+* *Envelope*: encrypted message with some metadata like topic and TTL echanged between Whisper nodes; a symmetric or assymetric key is needed to decrypt it and read the payload
+* *Offline message*: an expired envelope stored by a Whisper node permanently
+
+# Basic Assumption
+
+This protocol assumes that there is an Ethereum node that is capable of discoverying peers and implements Whisper V6 service. It also assumes that the participants of a given Whisper network accept messages with lowered PoW value.
```
mandrigin: I would use a numbered list of assumptions here to make them a bit more clear and stand-out.

Also it is useful to number stuff to refer later on, like (see 1.3.2.)

protocol.md
```diff
@@ -0,0 +1,196 @@
...
+
+The implementation of this protocol is mainly done in https://github.com/status-im/status-react and https://github.com/status-im/status-go repositories.
+
+The goal of this protocol is to allow people running Ethereum nodes with Whisper service enabled to exchange messages that are encrypted end-to-end in a way that guarantees darkness to some extent (messages are broadcasted through the network rather than send directly or using a known path).
+
+It's important to notice that messages are not limited to be text messages only. They can also have special meaning depending on the client implementation. For example, in the current implementation, there are message which informs about Eth requests or current user's location.
+
+This protocol consist of two components:
+* payload specification
+* Whisper adapter.
```
mandrigin: you also describe some offline messaging somewhere, isn't it a part of the protocol now?

protocol.md
```diff
@@ -0,0 +1,196 @@
...
+
+# Payload
+
+Payload is encrypted and decrypted using [transit format](https://github.com/cognitect/transit-format). It is a text-based format so that each decrypted message's payload can be easily read.
+
+Payload contains the following fields:
+* text `string`
+* content type `enum { text/plain }`
+* message type `enum { public-group-user-message, user-message }`
+* clock `int64`
```
mandrigin: difference between clock and timestamps should be stated

protocol.md
mandrigin: One more assumption: time is in sync for all the nodes in Status Whisper Network.

protocol.md
```diff
@@ -0,0 +1,196 @@
...
+* 1: array which items correspond to the described payload fields above
+
+For more details regarding serialization and deserialization please consult [transit format](https://github.com/cognitect/transit-format) specification.
+
+## Content types
+
+Here is a list of supported content types:
+* `text/plain` identifies a message which content is a plain text
+
+TODO: are there other types? How is it useful? Sending Eth requests?
```
mandrigin: maybe @yenda or @rasom know?

protocol.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Quoting
+
+TBD
+
+# Whisper adapter
+
+Whisper has been chosen as an messages exchange protocol because it was designed as an off-chain communcation layer for the Ethereum nodes. It supports e2e encryption and uses epidemic spread to route data to all members of the network. It also provides [darkness to some extent](https://github.com/ethereum/go-ethereum/wiki/Achieving-Darkness).
+
+However, Whisper was not designed to handle huge number of messages and real-time communication. These are the tradeoffs that we accepted when implementing this protocol.
```
mandrigin: It would be useful to point, which tradeoffs are there exactly.

protocol.md
```diff
@@ -0,0 +1,196 @@
...
+
+This protocol can either work using a Whisper service which requires a protocol implementation to run in the same process as a Whisper node as well as using a Whisper client which might run as a separate Ethereum node and communicate through IPC or WebSocket.
+
+There is some tight coupling between the payload and Whisper:
+* Whisper message topic depends on the actual message type
```
mandrigin: I would go so far that I would have a separate entries about our usage of topics and encryption in Whisper. It is an important topic.

protocol.md
```diff
@@ -0,0 +1,196 @@
...
+## Whisper node configuration
+
+If you want to run a Whisper node and receive messages from Status clients, you need to configure it properly, otherwise, some message might not be received.
+
+Whisper's Proof Of Work algorithm is used to to deter denial of service and various spam/flood attacks against the Whisper network. The sender of a message must perform some work which in this case means performing some computations. Because Status main client is a mobile client, this easily leads to battery draining and poor performance of the app itself. Hence, these clients send envelopes with the following settings:
+* proof-of-work equal `0.002`
+* proof-of-work time equal to `10`
+* time-to-live equal `10`
+
+If you want to receive messages from a mobile client, you need to run Whisper node with PoW set to `0.002`. In case of `geth`, this option can be overriden with `-shh.pow=0.002` flag.
```
mandrigin: hmm, maybe there can be a full example how a stock geth can be a Status node? like command-like `geth -ssh -ssh.pow=0.002`, etc

protocol.md
mandrigin: Okay, I see you have adressed it later, so maybe a note like "addressed later in the document" or "see chapter 3.2.1." might be useful.

protocol.md
```diff
@@ -0,0 +1,196 @@
+Status Secure And Decentralized Messaging Protocol
+==================================================
+
```
mandrigin, line 13: Table of contents would be very useful here.

mandrigin (commented):
Awesome progress with this document!
I left some initial feedback.

protocol.md
adambabik: I don't mention fleet below in the spec I believe so that's why I didn't put it in there.

protocol.md
adambabik: Yes, I believe I put a header for that to explain but if not I will do it.

protocol.md
adambabik: Actually that might be a good distinction. I put it under Whisper adapter but it could be a separate point as well.

protocol.md
adambabik: I can rewrite this part but we accepted tradeoffs that Whisper can't handle huge amount of concurrent messages and is not designed for real-time communicate, i.e. there is a latency often > 1-2s.

protocol.md
adambabik: This or adding an appendix that would provide more information how to run geth node with the protocol required values.

oskarth:
Asking @decanus to review to get a set of fresh eyes

Litmus test: (a) Can implement a Status client with these docs (b) Understanding of what this means in terms of security properties/guarantees etc (or at least understanding of gaps, e.g. anonymity not quantified etc)

protocol.md
```diff
@@ -0,0 +1,196 @@
...
+## Encryption
+
+Protocol distinguishes messages encrypted using asymmetric and symmetric encryptions.
+
+Symmetric keys are created using `shh_generateSymKeyFromPassword` Whisper JSON-RPC method which accepts one param, a string.
+
+Messages encrypted with asymmetric encryption should be encrypted using recipeint's public key so that only the recipient can decrypt them.
+
+# Perfect Forward Secrecy (PFS)
```
oskarth, line 231: Can fold in this https://github.com/status-im/status.im/blob/556217b63e77a38e8a7bfe9f2b748deacadd4fec/source/research/pfs.md

protocol.md
```diff
@@ -0,0 +1,196 @@
...
+
+TBD
+
+# Device syncing
```
oskarth, line 248: Perhaps @cammellos can assist on this?

protocol.md
```diff
@@ -0,0 +1,196 @@
...
+
+# Public messages
+
+TODO: describe how to send a public message starting from adding a key in Whisper etc.TBD
+
+# Group messages
+
+TODO: describe how to send a group message starting from adding a key in Whisper etc.
+
+# Offline messages
```
oskarth, line 331: Mention requirement to be HA

oskarth: Mention trustedPeer and implications for connection (direct TCP, anonymity ~ and no pow/ttl respect)

protocol.md
```diff
@@ -0,0 +1,196 @@
...
+
+TBD
+
+## Topic
+
+There are two types of Whisper topics the protocol uses:
+* static topic for private chats
+* dynamic topic based on a chat name for public chats
+
+Static topic is always the same and its hex representation is `0xf8946aac`. It is also called a discovery topic. Having only one topic for all private chats has an advantage as it's very hard to figure out who talks to who. A drawback is that everyone receives everyone's messages but they can decrypt only these they have private keys for.
```
oskarth: call out discovery topic as a special term, e.g. Define Constant, possibly earlier in doc

similar to calling out specific encryption algos/curves used

protocol.md
```diff
@@ -0,0 +1,196 @@
...
+Notice: this protocol is documented post factum. The goal of it is to clearly present the current design and prepare the ground for its second version.
+
+The implementation of this protocol is mainly done in https://github.com/status-im/status-react and https://github.com/status-im/status-go repositories.
+
+The goal of this protocol is to allow people running Ethereum nodes with Whisper service enabled to exchange messages that are encrypted end-to-end in a way that guarantees darkness to some extent (messages are broadcasted through the network rather than send directly or using a known path).
+
+It's important to notice that messages are not limited to be text messages only. They can also have special meaning depending on the client implementation. For example, in the current implementation, there are message which informs about Eth requests or current user's location.
+
+This protocol consist of two components:
+* payload specification
```
oskarth: Suggest using *italics* for specific meaningful terms, might also want to use common terminology like "wire protocol", not sure

protocol.md
oskarth, line 52: Somewhere it should mention Geth as req due to Parity incompatiblity (with link to implementation in Parity and Nimbus)

oskarth (commented):
Left some comments, good start

protocol.md
```diff
@@ -0,0 +1,196 @@
...
+
+This protocol assumes that there is an Ethereum node that is capable of discoverying peers and implements Whisper V6 service. It also assumes that the participants of a given Whisper network accept messages with lowered PoW value.
+
+# Protocol Overview
+
+Notice: this protocol is documented post factum. The goal of it is to clearly present the current design and prepare the ground for its second version.
+
+The implementation of this protocol is mainly done in https://github.com/status-im/status-react and https://github.com/status-im/status-go repositories.
+
+The goal of this protocol is to allow people running Ethereum nodes with Whisper service enabled to exchange messages that are encrypted end-to-end in a way that guarantees darkness to some extent (messages are broadcasted through the network rather than send directly or using a known path).
```
decanus: `sent directly` rather than `send directly`

protocol.md
decanus: `discovering` not `discoverying`

decanus (changes requested):

cammellos:
@adambabik I have updated pfs docs, pr is still pending but you can already link the page https://status.im/research/pfs.html https://github.com/status-im/status.im/pull/267

protocol.md
mandrigin: I'd prefer to have some info here too, it is very useful when skimming through the document, so I don't have to jump around too much.

protocol.md
mandrigin: yeah, I would explicitly state these ones in the doc, they are useful to know.

protocol.md
```diff
@@ -0,0 +1,249 @@
...
+
+TODO: are there other types? How is it useful? Sending Eth requests?
+
+## Message types
+
+Here is a list of supported message types:
+* `public-group-user-message` is a message to the public group
+* `user-message` is a private message.
+
+TODO: how is it useful? What are other supported message types?
```
adambabik: @yenda @rasom can you guys answer this?

protocol.md
```diff
@@ -0,0 +1,249 @@
...
+
+`timestamp` is Unix time calculated when the message is created (TODO: does it come from Whisper or the system?)
+
+`clock` is calculated using the algorithm of [Lamport timestamps](https://en.wikipedia.org/wiki/Lamport_timestamps). When there are messages available in a chat, `clock`'s value is calculated as `last-message-clock-value + 1`. If there are no messages, `clock` is initialized with `timestamp`'s value.
+
+`clock` value is used for the message ordering. Due to the used algorithm, distributed nature of the system and physics, we achieve casual ordering which might produce counterintuitive results in some edge cases. For example, when one joins a public chat and sends a message before receiving the exist messages, their message `clock` value might be lower and the message will end up in the past when the historical messages are fetched.
+
+## Quoting
+
+TBD
```
adambabik, line 147: @yenda @rasom and this one.

protocol.md
mandrigin: I'd explained it here anyway, fleets are a part of our messaging infrastructure, just like mailservers.

oskarth:
Can we fold in the PFS doc here? Fine to duplicate it in docs repo, but this document should be self-contained

oskarth:
@adambabik can we ensure this document follows https://tools.ietf.org/html/rfc2119 standard? I.e. use MUST etc capitalized to indicate contract spec follows more precisely.

adambabik:
@oskarth totally, the last two days I worked on confirming JSON-RPC service is easy to implement and did not put new thing in this specification but I want to finish the gaps related to how to send/receive messages and then we can revise it and follow the standard.

protocol.md
```diff
@@ -0,0 +1,249 @@
...
+Whisper's Proof Of Work algorithm is used to to deter denial of service and various spam/flood attacks against the Whisper network. The sender of a message must perform some work which in this case means performing some computations. Because Status main client is a mobile client, this easily leads to battery draining and poor performance of the app itself. Hence, these clients send envelopes with the following settings:
+* proof-of-work equal `0.002`
+* proof-of-work time equal to `10`
+* time-to-live equal `10`
+
+If you want to receive messages from a mobile client, you need to run Whisper node with PoW set to `0.002`. In case of `geth`, this option can be overriden with `-shh.pow=0.002` flag.
+
+TODO: provide an instruction how to start a Whisper node with proper configuration using geth.
+
+## Keys management
```
oskarth, line 174: @corpetty could you help out here wrt key paths etc? Thinking re keycard too

corpetty: yeah, Keycard def has the most up to date info on what the changes will be when keycard is integrated (separate keys).  I'll follow up with them and make sure the schematics I have for it are up to date.

protocol.md
corpetty: I tend to agree with @mandrigin here.  If we use it and we rely on it, we should describe it.

protocol.md
```diff
@@ -0,0 +1,249 @@
...
+It's important to notice that messages are not limited to be text messages only. They can also have special meaning depending on the client implementation. For example, in the current implementation, there are message which informs about Eth requests or current user's location.
+
+This protocol consist of three components:
+* payload specification
+* Whisper adapter
+* Offline messaging.
+
+The payload specification describes how the messages are encoded and decoded and what each fields mean. This is required to properly interpret messages by the client.
+
+Whisper adapter specifies interaction with the Whiper service with regards to keys management, configuration and some metadata (like topic) required to properly process and encrypt/decrypt messages.
```
corpetty: mispelled Whisper here.

protocol.md
```diff
@@ -0,0 +1,249 @@
...
+
+`clock` value is used for the message ordering. Due to the used algorithm, distributed nature of the system and physics, we achieve casual ordering which might produce counterintuitive results in some edge cases. For example, when one joins a public chat and sends a message before receiving the exist messages, their message `clock` value might be lower and the message will end up in the past when the historical messages are fetched.
+
+## Quoting
+
+TBD
+
+# Whisper adapter
+
+Whisper has been chosen as an messages exchange protocol because it was designed as an off-chain communcation layer for the Ethereum nodes. It supports e2e encryption and uses epidemic spread to route data to all members of the network. It also provides [darkness to some extent](https://github.com/ethereum/go-ethereum/wiki/Achieving-Darkness).
```
corpetty: `an` should be `a`.

protocol.md
corpetty: Also, why these trade-offs exist fundamentally... if not here, then a link to them.

protocol.md
```diff
@@ -0,0 +1,249 @@
...
+
+# Public messages
+
+TODO: describe how to send a public message starting from adding a key in Whisper etc.TBD
+
+# Group messages
+
+TODO: describe how to send a group message starting from adding a key in Whisper etc.
+
+# Offline messages
```
corpetty, line 331: Maybe this section can be changed to be just a "whisper node heirarchy" that details the specifics of the different types of clients that exist on the network, their particular use-case, load, and required trust.
- mobile
- whisper node
- mailserver
- ULC and LES ??

corpetty (commented):
Great work, I'll start working on the key paths

protocol.md
adambabik: But we should describe the current flow here. I am not sure when the keycard things is gonna happen, also, does it gonna change anything for the current users who do not have keycard?

protocol.md
adambabik: I am not sure we want to do that. I like the idea that each Whisper node is the same, however, in fact we have light Whisper clients and mail servers so it makes sense. I would keep this section but create another one which explains these slight differences between various nodes. Does that sound good?

protocol.md
adambabik: Well, it's just our internal way of naming a group of servers. Technically each fleet is a separate Whisper network and this is self explanatory. I don't see a point in introducing a term which can be replaced with something much clearer which is not specific to our implementation. Again, this is protocol spec, it should be separate from the implementation.

protocol.md
adambabik: FYI updated this section.

protocol.md
yenda: Are they even really isolated networks? Wouldn't a peer with connectivity on two fleets make them share all messages (unless they are running incompatible versions)

protocol.md
yenda: There is many other types:
```clojure
(def content-type-text "text/plain")
(def content-type-sticker "sticker")
(def content-type-status "status")
(def content-type-command "command")
(def content-type-command-request "command-request")
(def content-type-emoji "emoji")
```

yenda: It is useful at the UI level to render messages differently based on their content type

protocol.md
yenda: There is 3 messages types allowed in the schemas
```clojure
(spec/def ::message-type #{:group-user-message :public-group-user-message :user-message})
```

it is useful to distinguish public and private group and 1-1 chats. There is probably better solutions but it would break compatibility

protocol.md
yenda: I don't understand what is suppose to be put there? is it about the replies?

protocol.md
```diff
@@ -0,0 +1,338 @@
...
+
+Protocol distinguishes messages encrypted using asymmetric and symmetric encryptions.
+
+Symmetric keys are created using `shh_generateSymKeyFromPassword` Whisper JSON-RPC method which accepts one param, a string.
+
+Messages encrypted with asymmetric encryption should be encrypted using recipient's public key so that only the recipient can decrypt them.
+
+# Perfect Forward Secrecy (PFS)
+
+TODO: link to a separate document (currently in the PR).
```
oskarth, line 244: I think we want this inline

protocol.md
```diff
@@ -0,0 +1,338 @@
...
+
+# Abstract
+
+Ethereum empowers users and developers to interact with totally new kind of applications called Dapps (Decentralized applications). These application allows to interact with the blockchain on a completely new level which is not only about exchanging values but also executing arbitrary logic. This logic can form very sophisticated programs like DAOs (Decentralized autonomous organizations). The missing part here is how users of Dapps can communicate securely and in a decentralized way with each other. Communication is an essential part of any activity. In this document, we specify a secure and decentralized messaging protocol that is capable of running on the Ethereum network.
+
+# Introduction
+
+TBD
+
+# Terminology
```
oskarth, line 49: Might want to define a section on crypto primitives etc, there's a section on this in PFS afaik

protocol.md
```diff
@@ -0,0 +1,338 @@
...
+* *Message*: decrypted Whisper message
+* *Envelope*: encrypted message with some metadata like topic and TTL sent between Whisper nodes; a symmetric or asymmetric key is needed to decrypt it and read the payload
+* *Offline message*: an expired envelope stored by a Whisper node permanently
+
+# Basic Assumption
+
+This protocol assumes the following:
+1. There is an Ethereum node that is capable of discovering peers and implements Whisper V6 service.
+2. Participants of a given Whisper network in order to communicate with each other accept messages with lowered PoW value.
+3. Time is synced between all nodes participating in the given network (this is intrinsic Whisper requirement).
```
oskarth: This seems false, what does this mean more precisely?

protocol.md
oskarth: Think so

protocol.md
oskarth: Anonymity trilemma, though worth noting this hasn't been studied rigorously by any academics etc and it's largely a theoretical claim

protocol.md
```diff
@@ -0,0 +1,338 @@
...
+
+Whisper has been chosen as an messages exchange protocol because it was designed as an off-chain communication layer for the Ethereum nodes. It supports e2e encryption and uses epidemic spread to route data to all members of the network. It also provides [darkness to some extent](https://github.com/ethereum/go-ethereum/wiki/Achieving-Darkness).
+
+However, Whisper was not designed to handle huge number of messages and real-time communication. These are the tradeoffs that we accepted when implementing this protocol.
+
+This protocol can either work using a Whisper service which requires a protocol implementation to run in the same process as a Whisper node as well as using a Whisper client which might run as a separate Ethereum node and communicate through IPC or WebSocket.
+
+There is some tight coupling between the payload and Whisper:
+* Whisper message topic depends on the actual message type (see [Topic](#topic))
+* Whisper message uses a different key (asymmetric or symmetric) depending on the actual message type (see [Keys management](#keys-management))
```
oskarth, line 162: This should refer back to more precise crypto things used

oskarth (changes requested):
Do you want some comments as commits to this PR btw @adambabik or will this mess up your workflow?

protocol.md
mandrigin: It isn't false, Whisper uses timestamps and PoW to calculate TTLs, so if you timestamp is either too far away in the past or in the future it will get discarded, and moreover, your node will get dropped.

mandrigin: That's why we even introduced NTP sync in status-go, but it doesn't work immediately, it takes some time to get the responses, and if the time of your local device is off, say by 2 minutes, you won't be able to participate in the network.

oskarth: NTP != absolute time sync (impossible unless you have atomic time and bounded by some absolute term to XX probability), so we can phrase it more accurately in terms of dependency on wall clock time.

protocol.md
adambabik: Yes, it would but in this case they just stop being separate networks in the discussion. I don't see it being important for the protocol itself. It is agnostic how the network is formed. What I can do is to introduce `Whisper network` term.

protocol.md
mandrigin: okay, so something like "difference between local clocks of nodes is within 30 seconds?"

cammellos: I guess it should be phrased more accurately, it might give the impression that time is *actively* synced among participants.

`Two peers in a whisper network will only be able to communicate if their relative clock drift is within 20s, otherwise messages will be discarded`

An in the implementation section we can mention NTP, although that's not part of the protocol, and any mean (or none) to get your clock within those 20s will do.

protocol.md
adambabik: @yenda thanks! Can you please also provide some information what is each of these types?

Also @oskarth @mandrigin, the question here is whether we should support a strict list of content types or select a few as required ones and mention that other content types can be added by specific implementations?

protocol.md
mandrigin: they aren't just discarded, you (as a peer) are dropped too, as soon as you send an outdated message (or a message from the future).

protocol.md
adambabik: Yes, it's how replies are handled.

protocol.md
adambabik: Then I would add it after this is merged. It's already pretty long. I will put a note to inline PFS document here. It probably needs to be adjusted a bit anyway.

protocol.md
adambabik: Question if it's needed. The protocol itself relies on Whisper spec entirely and it's kind of out of our control. We should link to Whisper spec.

protocol.md
oskarth: Hm sure, not sure best way to do this. It's a bit app specific, so kind of like extensions. I.e. stickers might not work on CLI.  Maybe phrase as MAY support and SHOULD textplain?

protocol.md
oskarth: Yeah you are probably right

protocol.md
oskarth: Whisper spec is not good enough imo, and even so we want to at least abbreviate. I.e. Double Ratchet spec good enough but still take gist of it (primitives and what guarantees it provides, and some on flow)

oskarth:
@adambabik Agree this PR is getting to be quite huge. Perhaps it would make sense to merge a draft wit comments inline and then reopen a new one?

protocol.md
adambabik: I would link to http://eips.ethereum.org/EIPS/eip-627. I can't find anything more detailed in Whisper docs. We don't control cryptographic algorithms here.

oskarth (commented):
OK merging as draft assuming comments here are captured as TODOs in doc and we create a new one

adambabik:
@oskarth yeah, let's do it. As many people has started commenting today, let's make today until 3pm CET the final "comment day". What's not resolved, I will put as `TODO`s and merge this PR later today. Next, we will open very small PRs addressing each `TODO`.

protocol.md
```diff
@@ -0,0 +1,338 @@
+Status Secure And Decentralized Messaging Protocol
```
oskarth: See SIP-1 PR for rough draft, can add metadata and link to README if you want to merge, even while we refine more precise format

protocol.md
oskarth: That's because the Whisper spec isn't good enough. It should be possible to implement e.g. a client in Nim from this, and if there's a problem with some primitive we use (say hash fn, or slow KDF or whatever) it should be very clear what we are relying on. E.g. see https://code.briarproject.org/briar/briar-spec/blob/master/protocols/BTP.md and Signal/Whatsapp papers

mandrigin (approved):
I'm okay with merging it with TODOs

protocol.md
adambabik: 20 seconds is also purely geth's Whisper implementation thing. I will do something like this:
```
3. Time MUST be synced between all nodes participating in the given network (this is intrinsic requirement of the Whisper specification as well). A clock drift between two peers larger than 20 seconds MAY result in discarding incoming messages.
```

protocol.md
adambabik: In my opinion, before starting implementing this protocol, a client should have an implementation of Whisper V6 spec ready. We can discuss this later as it does not look like the most important part.

protocol.md
cammellos: `Time MUST be synced between all nodes participating` is too absolute I feel and gives you the impression that there's some syncing mechanism provided, I would only specify that `A clock drift between two peers larger than 20 seconds MAY result in discarding incoming messages` and that synchronization of clocks is left to each node and outside the scope of the protocol.

At the end of the day I believe the two important things to specify are:
1) A drift of more than x seconds might result in discarded incoming messages
2) Our protocol does not have any synchronization mechanism, and relies on external sources for that (NTP)

protocol.md
adambabik: We will address specific content types in the next PR. Some should be required but definitely not all of them.

protocol.md
adambabik: That's a fair point. I will change that.

PR #5, opened 2019-04-18
Sip 0001

oskarth:
Draft, adapted from EIP1. Lmk if people think this is a good idea

### DIFF
--- a/sips/sip-0001.md
+++ b/sips/sip-0001.md
@@ -1,0 +1,280 @@
+---
+eip: 1
+title: EIP Purpose and Guidelines
+status: Active
+type: Meta
+author: Martin Becze <mb@ethereum.org>, Hudson Jameson <hudson@ethereum.org>, and others
+        https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1.md
+created: 2015-10-27
+updated: 2015-12-07, 2016-02-01, 2018-03-21, 2018-05-29, 2018-10-17
+---
+
+## What is an EIP?
+
+EIP stands for Ethereum Improvement Proposal. An EIP is a design document providing information to the Ethereum community, or describing a new feature for Ethereum or its processes or environment. The EIP should provide a concise technical specification of the feature and a rationale for the feature. The EIP author is responsible for building consensus within the community and documenting dissenting opinions.
+
+## EIP Rationale
+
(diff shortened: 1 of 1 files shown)


sips/sip-0001.md
```diff
@@ -0,0 +1,239 @@
...
+
+It occasionally becomes necessary to transfer ownership of SIPs to a new champion. In general, we'd like to retain the original author as a co-author of the transferred SIP, but that's really up to the original author. A good reason to transfer ownership is because the original author no longer has the time or interest in updating it or following through with the SIP process, or has fallen off the face of the 'net (i.e. is unreachable or isn't responding to email). A bad reason to transfer ownership is because you don't agree with the direction of the SIP. We try to build consensus around an SIP, but if that's not possible, you can always submit a competing SIP.
+
+If you are interested in assuming ownership of an SIP, send a message asking to take over, addressed to both the original author and the SIP editor. If the original author doesn't respond to email in a timely manner, the SIP editor will make a unilateral decision (it's not like such decisions can't be reversed :)).
+
+## SIP Editors
+
+The current SIP editors are
+
+` TODO`
```
oskarth, line 196: How do we want to determine this?

sips/sip-0001.md
```diff
@@ -0,0 +1,239 @@
...
+
+SIP stands for Status Improvement Proposal. An SIP is a design document providing information to the Status community, or describing a new feature for Status or its processes or environment. The SIP should provide a concise technical specification of the feature and a rationale for the feature. The SIP author is responsible for building consensus within the community and documenting dissenting opinions.
+
+## SIP Rationale
+
+We intend SIPs to be the primary mechanisms for proposing new features, for collecting community technical input on an issue, and for documenting the design decisions that have gone into Status. Because the SIPs are maintained as text files in a versioned repository, their revision history is the historical record of the feature proposal.
+
+For Status implementers, SIPs are a convenient way to track the progress of their implementation. Ideally each implementation maintainer would list the SIPs that they have implemented. This will give end users a convenient way to know the current status of a given implementation or library.
+
+## SIP Types
```
oskarth, line 22: Which of these do we want to keep / start out with?

oskarth:
Another question: do we want to call it SIPs? Or have it be its own standalone thing outside of Status? What would that look like in this case?

Perhaps there are better/different role models than Python/Bittorrent/Bitcoin (reference client rather?)/Ethereum (yellow paper not EIP) here wrt specs, i.e. IETF etc in terms of decoupling protocol standards.

oskarth:
Ping for feedback ^ @status-im/status-protocol

oskarth:
Too heavy and unfinished state for now, closing.

PR #7, opened 2019-04-18
protocol: improve basic assumption clock drift note

adambabik:


oskarth (commented):
Is the 20 seconds a constant or how was this arrived at? Perhaps a note on why this happens (dropping envelopes/peers at Whisper layer wrt default TTL/PoW etc)

adambabik:
It is the implementation detail of geth's Whisper and it's constant. You can't configure it and originally it drops peers violating this. I think we changed it to discard such messages. Generally, in geth's Whisper implementation there are two thresholds for too old or too new (from the future) messages: (1) drop message, (2) drop peer. These thresholds are constant, non-configurable.

> Perhaps a note on why this happens (dropping envelopes/peers at Whisper layer wrt default TTL/PoW etc)

Question if we want to dig deep into these details. Generally, it is not directly related to PoW but as TTL affects PoW, it is affected indirectly. And generally, yes, having large TTL might help with avoid skewed clocks. Large TTL makes increases proof-of-work time though.

oskarth:
Should we merge this? But perhaps in the other upcoming doc, up to you @adambabik

PR #18, opened 2019-04-26
clean up Message Payload Specification (closes #15)

### LINKED ISSUE #15: Initial Message Payload Specification
This should give an overview of our message payload and contents. What's special about this one is that it doesn't really provide any security guarantees (or at least not many).

Things like upgradability are a concern here, as well as how to deal with new types (e..g what is a sticker in a CLI client?).

One of thinking of these is more like different clients, so TTT and Sticker market and Extensions would be their own specification. For now we can keep them listed in this initial spec, especially MUST ones like `text/plain`.

Also considerations re Transit, e.g. if we want to move to a different format like Protobuf. How would compatibility work.

Also considerations re what happens if someone puts a bad timestamp or clock.

See [x9.md](http://x9.md) for current draft.

## Acceptance criteria

* [ ] Someone can read this and implement a Status client (including links to other specs)
* [ ] They should clarify what security guarantees the protocol(s) provide and don't provide
* [ ] As well as what each protocol/layer requires and provides
* [ ] They should be described as orthogonal pieces, so if someone wants a different transport that should ideally require minimal tweaks to the protocols

In terms of who will judge, it'll be 2-3 main groups initially:

* [ ] Ourselves in protocol group
* [ ] Client implementers (Core, Embark and Nimbus people i.e. status-js and Stratus)
* [ ] Core dev calls participants

adambabik:


x8.md
```diff
@@ -1,34 +1,38 @@
...
 
-# Payload
+This specification describes how the payload of each message in the Status Protocol looks like. It does not care how the payload is encrypted or exchanged later.
 
 # Payload
 
-Payload is an array of bytes exchanged between peers. In the case of this protocol, the payload is encoded using [transit format](https://github.com/cognitect/transit-format). It is a text-based format so even encoded version is easily readable by humans. For more information about the payload in Status, please see [Initial Message Payload Specification Specification](x8.md).
-
-Payload contains the following fields:
+Payload is a struct (a compound data type) with the following fields:
```
oskarth: struct like In C or protobuf or what more precisely? I.e. if I want to implement a client in JS what do I need?

oskarth: Oops, nevermind. I got confused and forgot this was encoded in Transit, just didn't recognize the notion of structs from it :D

x8.md
```diff
@@ -41,160 +45,29 @@ As you can see, the message is an array and each index value has its meaning:
...
-4. clock `int64`
-5. timestamp `int64`
-6. content `struct { chat-id string, text string }`
-
-Example of a valid encoded payload:
-
-```
-["~#c4",["abc123","text/plain","~:public-group-user-message",154593077368201,1545930773682,["^ ","~:chat-id","testing-adamb","~:text","abc123"]]]
-```
+## Future encoding format
```
oskarth: What is this?

x8.md
```diff
@@ -1,34 +1,38 @@
...
-3. message type `enum { public-group-user-message, user-message, group-user-message }` (more in [Message types](#message-types))
+2. content type `enum` (more in [Content types](#content-types))
+3. message type `enum` (more in [Message types](#message-types))
 4. clock `int64`
 5. timestamp `int64`
 6. content `struct { chat-id string, text string }`
 
+# Encoding
+
+The payload is encoded using [Transit format](https://github.com/cognitect/transit-format). It is a text-based format so even encoded version is easily readable by humans.
```
oskarth: Something on JSON

Perhaps something on brief rationale (for @yenda and @cammellos maybe)

x8.md
```diff
@@ -41,160 +45,29 @@ As you can see, the message is an array and each index value has its meaning:
...
-
-```
-["~#c4",["abc123","text/plain","~:public-group-user-message",154593077368201,1545930773682,["^ ","~:chat-id","testing-adamb","~:text","abc123"]]]
-```
+## Future encoding format
 
-As you can see, the message is an array and each index value has its meaning:
-* 0: `c4` is a decoder handler identification for the current payload format. Identifications allow to register handlers for many different types of payload
-* 1: array which items correspond to the described payload fields above
+Although, Transit format suits for most of the Status Protocol needs, it has drawbacks:
```
oskarth: Why does it suit?

x8.md
```diff
@@ -41,160 +45,29 @@ As you can see, the message is an array and each index value has its meaning:
...
+* it is not widely popular and implementations for less popular languages might not be complete or exist at all
+* upgradability is a concern while, for example, protocol buffers are easily extensible.
```
oskarth: is this true or just about how we use it?

x8.md
```diff
@@ -41,160 +45,29 @@ As you can see, the message is an array and each index value has its meaning:
...
-## Content types
+# Content types
 
 Content types are required for a proper interpretation of incoming messaages. Not each message is a plain text but may carry a different information.
 
 The following content types MUST be supported:
-* `text/plain` identifies a message which content is a plain text
+* `text/plain` identifies a message which content is a plain text.
+
+There are also other content types used by the current implementations that MAY be implmeneted:
```
oskarth: typo

x8.md
```diff
@@ -203,230 +76,18 @@ The following messages types MUST be supported:
...
 * `group-user-message` is a message to the private group.
 
-## Clock vs Timestamp and message ordering
+# Clock vs Timestamp and message ordering
 
 `timestamp` MUST be Unix time calculated when the message is created. Because the peers in the Whisper network should have synchronized time, `timestamp` values should be fairly accurate among all Whisper network participants.
 
 `clock` SHOULD be calculated using the algorithm of [Lamport timestamps](https://en.wikipedia.org/wiki/Lamport_timestamps). When there are messages available in a chat, `clock`'s value is calculated based on the last received message in a particular chat: `last-message-clock-value + 1`. If there are no messages, `clock` is initialized with `timestamp`'s value.
 
 `clock` value is used for the message ordering. Due to the used algorithm and distributed nature of the system, we achieve casual ordering which might produce counterintuitive results in some edge cases. For example, when one joins a public chat and sends a message before receiving the exist messages, their message `clock` value might be lower and the message will end up in the past when the historical messages are fetched.
```
oskarth, line 78: What happens if clock or timestamp is deliberately set incorrectly?

(I understand a lot of these questions might not be able to answer straight away, so can just add as TODOs or so)

x8.md
```diff
@@ -1,34 +1,38 @@
...
+- [Encoding](#encoding)
+  - [Future encoding format](#future-encoding-format)
+- [Content types](#content-types)
+- [Message types](#message-types)
+- [Clock vs Timestamp and message ordering](#clock-vs-timestamp-and-message-ordering)
+- [Replies](#replies)
+- [Upgradability](#upgradability)
 
-# Payload
+This specification describes how the payload of each message in the Status Protocol looks like. It does not care how the payload is encrypted or exchanged later.
```
oskarth, line 19: Briefly on purpose, like to enable chat communication, things like stickers etc and uniform through many clients

Pointer to whitepaper here maybe? Enable SNT use cases outside of more conventional chat use cases

x8.md
cammellos: not sure how true this statement is, there might be some truth in it, but the differences between transit and protobuf are not that much, in terms of upgradability.

I know this sound trivial, but:

Very simply, protobuf is a map, Transit is an array, and that's really all there is to it.
You add new fields to a map by adding a key, you add fields in an array by appending a member.
You add a field to both, you don't break compatibility, you remove a field from either, you might break compatibility.

Is a map more upgradable then an array? Sure maps are more suited for these kind of things,
 and arrays make certain things harder (removing fields,for example, where you would want to keep positional order), but if you are removing fields, you likely are already breaking compatibility, so since adding is essentially identical and removing/changing is the only way to break compatibility, I would say that in terms of upgradability they are equivalent.

(there are are differences of course, one is typed the other isn't etc, but do not concern upgradability)

x8.md
cammellos: Low bandwidth I guess, compared to JSON, human readable (ish), I mean most of the data exchange formats would suit really, our use case is very simple.

x8.md
cammellos: Why not JSON you mean?
JSON would work just fine, just larger payload, I believe we had it before. I take transit was chosen mainly to reduce bandwitdth, but not sure.

oskarth: I mean it's "similar" but more efficient. Different from saying text based like IRC

x8.md
cammellos: If `clock` is deliberately set incorrectly your message can be  orddered arbitrarely (in the past or in the future), and that's unavoidable with logical clocks. To prevent this we discard messages that are "too" far off in the future, while we are ok on messages in the past (as they might come from a mailserver).

Timestamps currently is used as well for ordering (grouping by day), but that's a regression introduced during some performance optimization, which might result in messages not being casually ordered if the sender has it's clock drifted for more than 24 hours. Otherwise it is only used to show a timestamp next to each message.

x8.md
oskarth: Agree on similar to JSON + bandwidth + edn/Clojure connection (no way Go devs would've picked it I think as first choice)

x8.md
oskarth: Yeah makes sense, so from a protocol pov:

- note on it isn't cryptographically ordering events and this can be abused by bad actor
-  note on not relaying on timestamps

Then hints for app developers to:
- consider discarding messages too far in future (clock)
- possibly using timestamp for ordering/grouping per day with real clock (~)

cammellos: Not sure it makes sense to add what is not (it is not many things, it is only a few), and for the technical reader will know that lamport timestamps are just that, it's not a blockchain or a consensus based mechanism.

Also "cryptographically ordering events" is not a well defined term by any means, you might use cryptography to preserve ordering, but what is in your mind "cryptographical order"?

oskarth: It is a simple attack model and it weakens ability to order events in face of basic adversary. It might not matter a lot for chat,  but it definitely could for other applications like state channels etc.

Cryptographic hash function guarantees means you can maintain certain invariants which you can’t under Byzantine model (it is easy to come up with predictable integers)

oskarth: It’s an instance of don’t trust user input, especially in p2p

cammellos: I don't disagree we should add the fact that currently anyone can insert messages in the conversation, it's just a matter of how.

I would simply state the fact that currently we discard messages that have an higher clock value then `timestamp` + 1 month, while we accept any message that has a lower clock value than `timestamp` (this is the same as briar I believe, though briar uses a day).

The rest I don't think it belongs here, again, we should describe what it is and not what is not, and explain what are the attack vectors.

If we want to add something on the lines you proposed, it needs to be well defined at least:
i.e:
"cryptographical order" gives 4 results on google so I doubt people will understand what you mean ( I don't, best guess is that is the parent id stuff, or you are thinking about a blockchain like consensus model)

> Cryptographic hash function guarantees means you can maintain certain invariants 

list which invariants you  think are useful and to solve which problem, in our use case

> It’s an instance of don’t trust user input, especially in p2p

again, this kind of comments are not really useful, either you take the time to deal with the complexity and nuances, or they have the same usefulness as "don't break compatibility"

x8.md
yenda: Actually transit could have been using a map has well, the thing is the protocol was designed with upgradability in mind not forever backward and forward compatibility, the idea was not to extend the records but to use a new version and publish intermediary build that would support both.

So this is really not something that is desirable, just something we have to deal with for as long as we don't fix it by breaking compatibility once

x8.md
oskarth: > cryptographical order

This was sloppily written and is a shorthand for ordering messages like a DAG, based on message ids as dependencies that are cryptographic hashes. It proves that you have seen at least that message and can't arbitrarily claim you are from the future. Sorry if this wasn't clear. If it still isn't I'll add it in a separate PR next week.

> deal with the complexity and nuances, or they have the same usefulness as "don't break compatibility"

It's really not that complex and also not that big of a deal, it's just a caveat. It's important for people writing clients to be aware that they don't rely on it blindly. I disagree that general principles like "don't trust user input" and "don't break compatibility" aren't useful - I guess it depends on having a common understanding of these being goals and then working together to solve them. In this specific instance it's fairly straightforward as an enhancement (that may or may not be needed, depending on the application), whereas in other cases (compatibility and whisper topic partitioning) it might be more difficult and require more careful thinking, as you rightly point out. To me, it's useful to re-iterate (presumably shared) goals to remind ourselves of them and think about how they can be achieved, or what the limitations are.

cammellos: > It proves that you have seen at least that message and can't arbitrarily claim you are from the future.

This approach though brings other sort of issues, and it should fit our use case without changing the way things work now (should work on public chats, you should be able to post without having seen any message etc):

- difficult to order with partial information, i.e you have X1 -> X2 -> X4 -> X6 -> X8  X6 and X4 can't be ordered with a single parent-id, you can have multiple parent id but you are just making this less likely to happen

- you can still insert messages in history, unless you implement some form of consensus, or drop them

- you still need a meaningful way to break ties (multiple messages with the same parent id)

- can't post if you haven't fetched a message without some other mean of ordering etc

so again I would not add references to something that is not specified, unless there's a know implementation that people can reference ( which I am not aware of, closest is briar but only uses that only for threads).

Also having a higher/predictable clock value is not a big issue in our case, (you want your message to have the higher timestamp, other clients will do +1), we cap it just to allow messages from users who did not see any message to be ordered meaningfully (causality it is preserved even without this though, a parent-id method would likely need similar arrangement if we want to consider such cases).

>  I disagree that general principles like "don't trust user input" and "don't break compatibility" aren't useful

I might have worded this too strongly, I just feel that they provide little information, we all have enough working experience to know that we should not be blindly trusting user input, and we don't as a matter of fact, although we could be better. 
I would again just mention the tradeoffs (messages can be inserted in the past, high clock value case is already mentioned), and we can think about how to address those in a separate place.

x8.md
adambabik: Maybe wrong title but you mentioned that in the issue:
> Also considerations re Transit, e.g. if we want to move to a different format like Protobuf. How would compatibility work.

Personally, I would not talk about this here at all.

x8.md
adambabik: Reducing bandwidth sounds like a valid reason.

What's wrong with saying "text-based"? Both JSON and Transit are text-based. Unless you mean that it was not the reason of choosing it and this part suggests so.

x8.md
adambabik: It's not similar to JSON except for being text-based. It looks like low-bandwidth and Clojure-origin are the reasons then. I will change this.

x8.md
adambabik: > Very simply, protobuf is a map, Transit is an array, and that's really all there is to it.

In map the order is irrelevant while it is in an array. So, in transit I need to be careful to add the filed at the end, correct? In protobuf I specified field index.

Also, I was looking into that, and I couldn't find anything in the Transit repo that talks about upgradability while protobuf has a section dedicated to that: https://developers.google.com/protocol-buffers/docs/proto#updating.

So, are you for or against keeping this point?

The first statement is of course true, there is no good lib for Go, for example :D

cammellos: @adambabik correct, it's positional

for upgradability of transit, I believe is just because it's much simpler than protobuf (no types, no default values etc). Knowing the authors their thought is to discourage breaking changes, therefore choosing an array, which makes it awkward to remove fields, an encourage people to either append new fields, or use a completely different message type "tag".

In general I would word it probably like @yenda , the fault is not in transit, but mainly our own protocol was not thought out to be extended in the way we think of now.

x8.md
adambabik: > `clock` value is used for the message ordering.

will be

> `clock` value SHOULD is used for the message ordering.

Additionally, I will:
- add info about messages being inserted in the past,
- suggest that `timestamp` is recommended in the presentation layer,
- that this mechanism can be easily abused and clients MAY chose to drop messages too old or from the future.

cammellos: sounds good

> clock value SHOULD be used for the message ordering.

typo

`timestamp` is tricky, I personally think we should remove it, use whisper timestamps for the presentation layer (not sorting, not even grouping), and we could check against whisper timestamp/clock-value difference to handle some of the most obvious abuses

adambabik: That's a good point. I will add that for the presentation layer, Whisper envelope time should be used. In a separate PR (for better visibility and clean process of decision making), we can deprecate `timestamp` if we think it's not needed.

x8.md
yenda: yeah because the obvious answer is it couldn't work, the current protocol was never intended for backward and forward compatibility, so we either drag this on forever or at some point we'll break compatibility, which will certainly be the case with the new protocol anyway

yenda: While the current protocol had versioning, it was never used to maintain backward compatibility. Instead v1 of each message has been extended and as result there is no versioning and updates are ad-hoc with newer versions adding older fields to the message to maintain compatibility

x8.md
yenda: Yes the reason for transit over json was bandwidth and better clojure types supports, because we wanted to have an upgradable protocol at the time (not extensibility) we opted to use records, which save bandwidth as fields are known to both sides and therefore it is sent in an array. The record name was also shorthanded to represent message type and message type version. But as stated above this was shortcuted later on for extensibility and therefore there never used, so the array/record is completely useless and makes extensibility harder without a proper versioning scheme to keep track of changes

x8.md
yenda: correctm lower bandwidth, clojure data litterals support and extensibility

x8.md
yenda: Again, transit could use a map as well, right now it uses a record, which is a kinda typed array, the reason was upgradability, but because we traded it for extensibility, it is completely useless, and we lost track of versions

cammellos: oh I stand corrected about arrays in transit, I did not know that, thanks @yenda

adambabik:
@oskarth is it ok to merge? We can address missing parts in different PRs.

PR #34, opened 2019-07-17
Add v1 message wrapper

cammellos:
Adds the message wrapper from https://github.com/status-im/status-protocol-go/pull/19

I have also restructured `x8` as `Message` is only one type of messages exchanged by the app, there`s more that needs describing: `PairInstallation`, `SyncInstallation`, `ContactRequest`, `ContactUpdate` and a few more.

### DIFF
--- a/x8.md
+++ b/x8.md
@@ -8,27 +8,30 @@ created: 2019-04-22
 updated:
 ---
 
-- [Payload](#Payload)
+- [Wrapper](#Wrapper)
 - [Encoding](#Encoding)
-- [Content types](#Content-types)
-- [Message types](#Message-types)
-- [Clock vs Timestamp and message ordering](#Clock-vs-Timestamp-and-message-ordering)
-- [Replies](#Replies)
 - [Upgradability](#Upgradability)
 
 This specification describes how the payload of each message in the Status Protocol looks like. It does not care how the payload is encrypted or exchanged later.
 
 The payload must be flexible enough to support messenging but also cases described in [Status Whitepaper](https://status.im/whitepaper.pdf) as well as various clients created using vastly different technologies.
 
-# Payload
+# Wrapper
 
-Payload is a struct (a compound data type) with the following fields (order is importent):
-1. text `string`
-2. content type `enum` (more in [Content types](#content-types))
-3. message type `enum` (more in [Message types](#message-types))
-4. clock `int64`
-5. timestamp `int64`
-6. content `struct { chat-id string, text string }`
+Payloads are wrapped in a [protobuf record](https://developers.google.com/protocol-buffers/)
+record:
+
+```
+    message StatusProtocolMessage {
+      bytes signature = 1;
+      bytes payload = 2;
+    }
+```
+
+Where `signature` is the bytes of the signed `SHA3-256` of the payload, signed with
+the key of the author of the message.
+The signature is needed to validate authorship of the message, so that the message can be relayed to third parties.
+If a signature is not present but an author is provided by a layer below, the message is to be relayed to third parties and it's considered plausibly deniable.
 
 # Encoding
 
@@ -46,7 +49,32 @@ As you can see, the message is an array and each index value has its meaning:
 
 For more details regarding serialization and deserialization please consult [transit format](https://github.com/cognitect/transit-format) specification.
 
-# Content types
+# Message types
+
+- [Message](##Message)
+
+## Message
+
+The type `Message` represents a text message exchanged between clients.
+
+- [Payload](###Payload)
+- [Content types](###Content-types)
+- [Message types](###Message-types)
+- [Clock vs Timestamp and message ordering](###Clock-vs-Timestamp-and-message-ordering)
+- [Replies](###Replies)
+
+### Payload
+
+Payload is a struct (a compound data type) with the following fields (order is importent):
+1. text `string`
+2. content type `enum` (more in [Content types](#content-types))
+3. message type `enum` (more in [Message types](#message-types))
+4. clock `int64`
+5. timestamp `int64`
+6. content `struct { chat-id string, text string }`
+
+
+### Content types
 
 Content types are required for a proper interpretation of incoming messaages. Not each message is a plain text but may carry a different information.
 
@@ -60,7 +88,7 @@ There are also other content types that MAY be implemented by the client:
 * `command-request` TODO
 * `emoji` TODO
 
-# Message types
+### Message types
 
 Message types are required to decide how a particular message is encrypted (more in [Whisper > Message encryption](#message-encryption)) and what metadata needs to be attacjed (more in [Whisper > Topic](#topic)) when passing a message to the transport layer.
 
@@ -69,7 +97,7 @@ The following messages types MUST be supported:
 * `user-message` is a private message
 * `group-user-message` is a message to the private group.
 
-# Clock vs Timestamp and message ordering
+### Clock vs Timestamp and message ordering
 
 `timestamp` MUST be Unix time calculated when the message is created. Because the peers in the Whisper network should have synchronized time, `timestamp` values should be fairly accurate among all Whisper network participants.
 
@@ -77,7 +105,7 @@ The following messages types MUST be supported:
 
 `clock` value is used for the message ordering. Due to the used algorithm and distributed nature of the system, we achieve casual ordering which might produce counterintuitive results in some edge cases. For example, when one joins a public chat and sends a message before receiving the exist messages, their message `clock` value might be lower and the message will end up in the past when the historical messages are fetched.
 
-# Replies
+### Replies
 
 TODO
 


adambabik (approved):

adambabik:
Oh, can we add these not-yet-described messages into the spec as well with `TODO` annotations? Alternatively, we can create an issue.

x8.md
```diff
@@ -46,7 +47,32 @@ As you can see, the message is an array and each index value has its meaning:
...
+
+- [Payload](###Payload)
+- [Content types](###Content-types)
+- [Message types](###Message-types)
+- [Clock vs Timestamp and message ordering](###Clock-vs-Timestamp-and-message-ordering)
+- [Replies](###Replies)
+
+### Payload
+
+Payload is a struct (a compound data type) with the following fields (order is importent):
```
oskarth, line 68: typo: important

To clarify, this order ensures the hash is deterministic per client implementation above?

oskarth (approved):
perfect, one typo and question

x8.md
cammellos: no,  the hash is deterministic based of sha3 256, 
but the encoding is not.

This means that two different clients, encoding exactly the same message might result in different ids. That's because of protobuf non deterministic encoding (most of the time it is though , there s more on github discussion for protobuf).

In our case this is not a problem, as the hash is always calculated over the encoded version, and no client should decode/re-encode if it needs to propagate/resend, but should always use the unmodified received payload

decanus: @cammellos, that seems like an accident waiting to happen. I’d rather we pack fields separately so we always have the same hash and don’t hash the encoded message.

cammellos: what do you mean exactly by "packing fields separately" ?

oskarth (commented):
(pending approval after hash field order discussion)

x8.md
decanus: HASH(protofield1, ...); otherwise someone will fuck up. Someone always does.

oskarth: > This means that two different clients, encoding exactly the same message might result in different ids. That's because of protobuf non deterministic encoding (most of the time it is though , there s more on github discussion for protobuf).

I might be misjudging the problem, but I believe `**HASH("MESSAGE_ID", group_id, timestamp, body);**` from https://github.com/status-im/bigbrother-specs/blob/master/data_sync/mvds.md would be an example of that. Even though it is still using protobuf

oskarth (commented):
(silly github review interface)

x8.md
oskarth: I wonder how this would make sense for signatures over arbitrary data though, do we have any other example protocols to look at here?

Also cc @arnetheduck

cammellos: not sure you guys have got this right

oskarth: > Those unable to decrypt the message data are also unable to access the signature. The signature, if provided, is the ECDSA signature of the Keccak-256 hash of the unencrypted data using the secret key of the originator identity. The signature is serialised as the concatenation of the R, S and V parameters of the SECP-256k1 ECDSA signature, in that order. R and S are both big-endian encoded, fixed-width 256-bit unsigned. V is an 8-bit big-endian encoded, non-normalised and should be either 27 or 28.

From http://eips.ethereum.org/EIPS/eip-627

Seems to indicate field order of underlying data hash is calculated over is important, whereas above seems to indicate it is indeterministic.

As said above, could very well misunderstand something but it seems worthwhile to clarify at least

cammellos: signature should be over encoded data.

Otherwise it makes the protocol non upgradable as you can't add fields, as other clients will not parse that field, and won't use it in the signature.

that s a disaster that will happen.

oskarth what you are suggesting is no different from what i proposed, only we don't have a timestamp, messageid, groupid.


moreover the same problem that you are concerned about happen with that datasync code.

other protocols also calculate signatures over the encoded data for the same reasons (whisper, briar etc)

oskarth: > Those unable to decrypt the message data are also unable to access the signature. The signature, if provided, is the ECDSA signature of the Keccak-256 hash of the unencrypted data using the secret key of the originator identity. The signature is serialised as the concatenation of the R, S and V parameters of the SECP-256k1 ECDSA signature, in that order. R and S are both big-endian encoded, fixed-width 256-bit unsigned. V is an 8-bit big-endian encoded, non-normalised and should be either 27 or 28.

From http://eips.ethereum.org/EIPS/eip-627, where the data appears to be deterministic (encrypted version of envelopes, which are RLP encoded and afaik deterministic).

As said above, could very well misunderstand something but it seems worthwhile to clarify at least

oskarth: > signature should be over encoded data.
>
> Otherwise it makes the protocol non upgradable as you can't add fields, as other clients will not parse that field, and won't use it in the signature.

That's a fair point. I personally find it difficult to evaluate what the consequences are of not being able to get the same signature for the same data, but it could be that this is not something to worry about.

> moreover the same problem that you are concerned about happen with that datasync code.

Why? in MVDS spec it's over some bytes. If clients mess with order of those that's a different question, no?

oskarth: > no client should decode/re-encode if it needs to propagate/resend, but should always use the unmodified received payload

Isn't this adding assumptions on keeping certain payloads around? It at least seems like a trade-off to me, perhaps it's well established what the better solution is (don't care about order in encoding and sign the blob)

cammellos: Ok,
not sure what you are arguing for:

1) In the proposal Hash(protobuf encoded payload)
2) Suggested by dean Hash(Text, ClockValue)

The first is non deterministic because of protobuf, meaning if you encode the same content twice you may get a different hash. Can you make an example of where this is actually going to be a  problem? Just to understand what is the actual concern. The datasync example you sent is exactly this, but we only have a field here (as a matter of fact we should include the pk of the sender when calculating the messageID but that's another story).

The second is not even worth discussing, as any extra field will break the scheme, and works well only when you have a very well defined set of fields (whisper for example R,S, V , which won't change), but notice that it won't decode the data

We can argue that we should not be using protobuf because of this, but everyone seemed to be onboard and we all knew it was not deterministic.

Is any other proposal worth considering?

cammellos: > Isn't this adding assumptions on keeping certain payloads around? It at least seems like a trade-off to me, perhaps it's well established what the better solution is 

You can't decode/encode as fields that you don't know (say a field from a more recent version), will be silently discarded in most implementation, as can't be read.

That means that decoding is **lossy** so re-encoding is not an option.

oskarth: Not arguing for anything specific right now, just trying to explore the problem/solution space. I don't know. By default I do think your proposal is correct, but it still seems worth understanding and elaborating on the consequences a bit more.

I don't think being generally OK with protobuf as encoding necessarily means always using indeterministic ordering. In MVDS, you can see that the fields being hashed are individual protobuf fields, which means determinism is introduced. What's in the body is outside of the scope of that layer.

Do we have to rush this change? I'd like to hear Jacek's POV on this.

oskarth:
@cammellos how does validation of signature work? Also did you consider using a MAC?

x8.md
cammellos: > Why? in MVDS spec it's over some bytes. If clients mess with order of those that's a different question, no?

This is the same, the signature is over some bytes, not a specific format. Is on whatever the application layer passes. We could use transit, protobuf or anything, and this code will not change. At the moment we don't even use protobuf, we actually use transit, so it is deterministic. But it should not be a concern here, as the signature code does not have any understanding of what is the content of the data.

> In MVDS, you can see that the fields being hashed are individual protobuf fields, which means determinism is introduced.

This is exactly the same. It's a wrapper, it's not the **application layer** (otherwise it would not be a wrapper). It does not care what the application layer sends. There's no ordering here as there's only **1** field. Namely the payload. So with that respect is deterministic. 


The consequences I indicated is only if we use protobuf for encoding the messages, which currently we don't, we use transit, which afaik is deterministic.

oskarth: Good points, agree. It's probably due to the proximity of the wrapper and application layer that I personally mixed the two

cammellos:
@oskarth https://github.com/ethereum/go-ethereum/blob/d1aa605f1e8639769cdd75bcec3064c29a62b34a/crypto/secp256k1/secp256.go#L93
is the standard secp256 recovery from a signature and a signed payload.

> Also did you consider using a MAC?

We can't use macs, it only works if there's a shared secret, and this should apply for public messages as well, where we don't know how the recipient is.

1-to-1 messages can optionally be non-signed, as discussed in the discuss post. In that case we could use the MAC from the double ratchet, and those messages will be pd and can't be relayed.

oskarth:
Another thing: it might be useful to state more explicitly why we are introducing a signature, and what security goals this provides. As an example (https://crypto.stackexchange.com/questions/5646/what-are-the-differences-between-a-digital-signature-a-mac-and-a-hash), something like:

- integrity/authentication/non-repudiation
- which kind of keys it works for (vs e.g. Whisper underlying layer supporting both types)

As well a sentence or so on verification and requirements for it (keeping payload around)

What do you think?

cammellos:
makes sense, i will add that

oskarth:
> > Another thing: it might be useful to state more explicitly why we are introducing a signature, and what security goals this provides. As an example (https://crypto.stackexchange.com/questions/5646/what-are-the-differences-between-a-digital-signature-a-mac-and-a-hash), something like:
> 
> >     integrity/authentication/non-repudiation
> >     which kind of keys it works for (vs e.g. Whisper underlying layer supporting both types)
> > 
> > As well a sentence or so on verification and requirements for it (keeping payload around)
> > 
> > What do you think?
> > 
> 

> makes sense, i will add that


If we do this I'd be happy with this, @decanus?

cammellos:
@oskarth done, merging

PR #42, opened 2019-08-30
Provide details about peers discovery implementation in client

adambabik:


status-client-spec.md
```diff
@@ -109,10 +125,43 @@ v5](https://github.com/ethereum/devp2p/blob/master/discv5/discv5.md) and
...
-This part of the system is currently underspecified and requires further detail.
+A Status client MUST use at least one discovery method or use static nodes
+to communicate with other clients.
+
+Discovery V5 uses bootstrap nodes to discover other peers. Bootstrap nodes MUST support
+Discovery V5 protocol as well in order to provide peers. It is kademlia-based discovery mechanism
+and it might consume significant (at least on mobile) amount of network traffic to operate.
+
+In order to take advantage from simpler and more mobile-friendly peers discovery mechanism,
+i.e. Rendezvous protocol, one MUST provide a list of Rendezvous nodes which speak
```
oskarth, line 136: What is "one"? here? Is this as a bootstrap node?

If a bootstrap node doesn't, what happens? (Should this be a SHOULD?)

status-client-spec.md
```diff
@@ -109,10 +125,43 @@ v5](https://github.com/ethereum/devp2p/blob/master/discv5/discv5.md) and
...
+Rendezvous protocol. Rendezvous protocol is request-response discovery mechanism.
+It uses Ethereum Node Records (ENR) to report discovered peers.
+
+Both peers discovery mechanisms use topics to provide peers with certain capabilities.
+There is no point in returning peers that do not support a particular protocol.
+Status nodes that want to be discovered MUST register to Discovery V5 and/or Rendezvous
+with the `whisper` topic. Status nodes that are mail servers and want to
+be discoverable MUST additionally register with the `whispermail` topic.
```
oskarth, line 144: How does mailserver discovery work? I thought this was just as a normal Whisper node and that "discovery" was a manual process of adding static Whisper mailserver nodes

status-client-spec.md
```diff
@@ -109,10 +125,43 @@ v5](https://github.com/ethereum/devp2p/blob/master/discv5/discv5.md) and
...
+
+The recommended strategy is to use both mechanisms but at the same time implement a structure
+called `PeerPool`. `PeerPool` is responsible for maintaining an optimal number of peers.
+For mobile nodes, there is no significant advantage to have more than 2-3 peers and one mail server.
```
oskarth, line 148: Perhaps we should mention Eclipse attacks as a security consideration for this?

status-client-spec.md
```diff
@@ -109,10 +125,43 @@ v5](https://github.com/ethereum/devp2p/blob/master/discv5/discv5.md) and
 [modifications](https://github.com/status-im/rendezvous#differences-with-original-rendezvous)).
 Additionally, some static nodes MAY also be used.
 
-This part of the system is currently underspecified and requires further detail.
+A Status client MUST use at least one discovery method or use static nodes
+to communicate with other clients.
+
+Discovery V5 uses bootstrap nodes to discover other peers. Bootstrap nodes MUST support
```
oskarth, line 131: Can we generally clarify what the requirements are for each node type? I.e. as a bootstrap node, or as a mobile node etc.

oskarth (commented):
Thanks! Generally looks good, left some comments

oskarth:
Actually, since the diff is getting quite big, perhaps we should separate out things like discovery process etc into its own spec and link to it from the main spec? I want the main client spec to be quite concise and not go into too much detail.

@adambabik @decanus

status-client-spec.md
adambabik: "one" is a person operating a client in this case.

Rendezvous nodes are sort of bootstrap nodes but speaking the Rendezvous protocol. If someone wants to take advantage of Rendezvous, a list of Rendezvous nodes MUST be provided. The protocol itself does not need to be used though.

status-client-spec.md
adambabik: Mail servers can be discovered separately from Whisper nodes. Not all Whisper nodes are mail servers so if you'd like to discover only mail servers it should be possible. That's why we have separate topics.

A mail server node would usually register itself using both `whisper` and `whipermail` topics in the peers discovery protocol.

> I thought this was just as a normal Whisper node and that "discovery" was a manual process of adding static Whisper mailserver nodes

It works this way in our client now but there is nothing against discovering mail servers through the peers discovery algorithm. The problem is that you don't know whether you can trust these discovered mail servers. There was some work around that. cc @cammellos

status-client-spec.md
adambabik: Yes, it's a good point.

status-client-spec.md
adambabik: Yes but what kind of requirements would you like to distinguish? Protocols support, number of peers it seek to connect to or something more?

adambabik:
@oskarth I agree here. Peers discovery is a quite separate topic. Do we want to consider a basic use case where one has a node connected to a Whisper network and for more complicated use cases we provide links to e.g. how discover peers etc?

status-client-spec.md
oskarth: This should probably be clarified in the spec itself

status-client-spec.md
oskarth: > whisper and whipermail topics in the peers discovery protocol.

Ack

---

Can we add these details to the spec? Right now this is largely implicit knowledge

status-client-spec.md
oskarth: Any requirement that someone writing a client, or mailserver server, might care about. Things they MUST do SHOULD do or MAY do.

E.g. MUST support x version or above and SHOULD be connected to N nodes and MAY choose to do action x

oskarth:
> @oskarth I agree here. Peers discovery is a quite separate topic. Do we want to consider a basic use case where one has a node connected to a Whisper network and for more complicated use cases we provide links to e.g. how discover peers etc?

Both of these sounds like sections in a discovery spec. We should refer to other canonical specs as much as possible, assuming we are following their specs (discv5 discovery / rendezvous etc)


---

Merging this PR as is to have more of a common ground in the spec review session. However, all comments above still apply

status-client-spec.md
```diff
@@ -109,10 +125,43 @@ v5](https://github.com/ethereum/devp2p/blob/master/discv5/discv5.md) and
...
+Discovery V5 protocol as well in order to provide peers. It is kademlia-based discovery mechanism
```
arnetheduck, line 132: what's `discovery v5`?

PR #43, opened 2019-09-02
Describe chats concept

adambabik:
cc @cammellos


oskarth:
Merging to have common ground for spec review

PR #47, opened 2019-09-06
Minor fixes and typo corrections

kdeme:
I'm reading through the specs and I noticed some typos and some minor items that I believe to be incorrect or missing.

### DIFF
--- a/status-whisper-usage-spec.md
+++ b/status-whisper-usage-spec.md
@@ -284,7 +284,7 @@ Receiving public messages depends on Whisper filters idea. Upon receiving, messa
 
 1. Calculate a symmetric key using [`shh_generateSymKeyFromPassword`](https://github.com/ethereum/go-ethereum/wiki/Whisper-v6-RPC-API#shh_generatesymkeyfrompassword) passing public chat name as a string and save the result to `symKeyID`
 2. Call [`shh_subscribe`](https://github.com/ethereum/go-ethereum/wiki/Whisper-v6-RPC-API#shh_subscribe) with criteria:
-   1. `minPow` MAY be at least `0.002`
+   1. `minPow` MUST be `0.002` at most
    2. `topics` MUST be list of hex-encoded topics you expect messages to receive from (follow [Topic](#topic) section)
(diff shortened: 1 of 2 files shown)


oskarth:
Thanks @kdeme!

Added you to the repo too so you can request PR reviewers directly. Feel free to use `@status-im/status-protocol` to ping people for reviews too.

status-whisper-usage-spec.md
```diff
@@ -283,7 +284,7 @@ Receiving public messages depends on Whisper filters idea. Upon receiving, messa
 
 1. Calculate a symmetric key using [`shh_generateSymKeyFromPassword`](https://github.com/ethereum/go-ethereum/wiki/Whisper-v6-RPC-API#shh_generatesymkeyfrompassword) passing public chat name as a string and save the result to `symKeyID`
 2. Call [`shh_subscribe`](https://github.com/ethereum/go-ethereum/wiki/Whisper-v6-RPC-API#shh_subscribe) with criteria:
-   1. `minPow` MUST be at least `0.002`
+   1. `minPow` MAY be at least `0.002`
```
oskarth: @adambabik can you confirm if this and above statements are accurate?

adambabik: If `minPow` is above `0.002`, messages sent by the current clients won't be received.
```suggestion
   1. `minPow` MUST be `0.002` at most
```
is the correct version then.

Unless one wants to implement the spec but use higher PoW across their clients.

kdeme: Yeah, that would be best (as you state that there is no intention of changing the PoW per chat or client or so).

Only reason then why you could still put MAY is that you can actually just not set a minPow at all. So any pow will be accepted and the filtering is done by the node PoW requirement. Not the filter pow. But that might just confuse people.

adambabik: But then we would need to say that one MUST use `0.002` at most in the node PoW requirement. I agree that it would be confusing.

oskarth: @kdeme do you want to amend and we can merge this?

PR #49, opened 2019-09-09
Add payload specs

cammellos:
Add payload specs describing all kind of currently used messages

### DIFF
--- a/status-group-chats-spec.md
+++ b/status-group-chats-spec.md
@@ -34,7 +34,7 @@ The format is:
 
 ```
 {
-  "events": [],
+  "events": [struct {"type": string, "member": string, "members": [string], "clock-value": uint, "name": string],
   "signature": string,
   "chat-id": string
 }
@@ -43,11 +43,12 @@ The format is:
 ### Chat ID
 
 Each membership update MUST be sent with a corresponding `chat-id`. 
-The format of this chat id MUST be a string, [UUID](https://tools.ietf.org/html/rfc4122 ), concatenated with the hex-encoded public key of the creator of the chat. This chat-id MUST be validated by all clients, and should be discarded if it does not follow these rules.
+The format of this chat id MUST be a string, [UUID](https://tools.ietf.org/html/rfc4122 ), concatenated with the hex-encoded public key of the creator of the chat. This chat-id MUST be validated by all clients, and MUST be discarded if it does not follow these rules.
 
 ### Signature
 
-The signature for each event is calculated by creating a JSON array of all the `events` sorted by `clock-value` in ascending order, where each event is transformed in an array of tuples `field-name`, `value`. The last element of the array MUST be the `chat-id`. 
+The signature for each event is calculated by creating a JSON array of all the `events` sorted by `clock-value` in ascending order, where each event is transformed in an array of tuples `field-name`, `value`, sorted by `field-name` in ascending alphabetical order. The last element of the array MUST be the `chat-id`. 
+Empty fields MUST be removed.
 
 For example the event:
 
@@ -89,12 +90,13 @@ Results in the structure:
    ]
 ```
 
-This structure is then stringified and the `Keccak256` of the string is then signed using its private key by the author and added to the payload.
+This structure is then stringified collapsing all whitespaces and the `Keccak256` of the string is then signed using its private key by the author and added to the payload.
       
 
 ### Group membership event
 
-Any group membership event receive MUST be verified using by calculating the signature as in the method described above, and the author MUST be extracted from it, if the verification fails the event MUST be discarded.
+Any group membership event received MUST be verified by calculating the signature as per the method described above. 
+The author MUST be extracted from it, if the verification fails the event MUST be discarded.
 
 #### chat-created
 
@@ -137,7 +139,7 @@ If the event is valid the chat name SHOULD be changed to `name`.
 
 A members added event is used by admins to add members to the chat.
 Upon receiving this event a client MUST validate the `chat-id` provided with the updates and MUST ensure the author of the event is an admin of the chat, otherwise the event MUST be ignored.
-If the event is valid a client SHOULD update the list of members of the chat who have not joined, adding the `members` received.
+If the event is valid a client MUST update the list of members of the chat who have not joined, adding the `members` received.
 `members` is an array of hex encoded public keys.
 
 #### member-joined
@@ -152,7 +154,7 @@ If the event is valid a client SHOULD update the list of members of the chat who
 
 A members joined event is used by a member of the chat to signal that they want to start receiving messages from this chat.
 Upon receiving this event a client MUST validate the `chat-id` provided with the updates and MUST ensure the author of the event is the same as the one specified by the `member` field.
-If the event is valid a client SHOULD update the list of members of the chat who joined, adding `member`. Any `message` sent to the group chat should now include the newly joined member.
+If the event is valid a client MUST update the list of members of the chat who joined, adding `member`. Any `message` sent to the group chat should now include the newly joined member.
 
 #### admins-added
 
@@ -166,7 +168,7 @@ If the event is valid a client SHOULD update the list of members of the chat who
 
 An admins added event is used by admins to add make other admins in the chat.
 Upon receiving this event a client MUST validate the `chat-id` provided with the updates, MUST ensure the author of the event is an admin of the chat and MUST ensure all `members` are already `members` of the chat, otherwise the event MUST be ignored.
-If the event is valid a client SHOULD update the list of admins of the chat, adding the `members` received.
+If the event is valid a client MUST update the list of admins of the chat, adding the `members` received.
 `members` is an array of hex encoded public keys.
 
 #### member-removed
@@ -184,7 +186,7 @@ Upon receiving this event a client MUST validate the `chat-id` provided with the
 - If the author of the event is an admin, target can only be themselves or a non-admin member.
 - If the author of the event is not an admin, the target of the event can only be themselves.
 -
-If the event is valid a client SHOULD remove the member from the list of `members`/`admins` of the chat, and no further message should be sent to them.
+If the event is valid a client MUST remove the member from the list of `members`/`admins` of the chat, and no further message should be sent to them.
 
 #### admin-removed
 
@@ -199,4 +201,4 @@ If the event is valid a client SHOULD remove the member from the list of `member
 An admin-removed event is used to drop admin privileges.
 Upon receiving this event a client MUST validate the `chat-id` provided with the updates, MUST ensure that the author of the event is also the target of the event.
 
-If the event is valid a client SHOULD remove the member from the list of `admins` of the chat.
+If the event is valid a client MUST remove the member from the list of `admins` of the chat.
--- a/status-group-chats-spec.md
+++ b/status-group-chats-spec.md
@@ -1,0 +1,1 @@
+# Group chat specs
--- a/status-group-chats-spec.md
+++ b/status-group-chats-spec.md
@@ -1,1 +1,202 @@
-# Group chat specs
+# Status Group Chat Specification
+
+> Version: 0.1 (Draft)
+>
+> Authors: Andrea Maria Piana <andreap@status.im>
+>
+
+
+## Table of Contents
+
+- [Abstract](#abstract)
+- [Membership updates](#membership-updates)
+  - [Chat ID](#chat-id)
+  - [Signature](#signature)
+  - [Group membership event](#group-membership-event)
+    - [chat-created](#chat-created)
+    - [name-changed](#name-changed)
+    - [members-added](#members-added)
+    - [members-joined](#members-joined)
+    - [admins-added](#admins-added)
+    - [members-removed](#members-removed)
+    - [admin-removed](#admin-removed)
+
+
+## Abstract
+
+This documents describes the group chat protocol used by the status application. Pairwise encryption is used among member so a message is exchanged between each participants, similarly to a one-to-one message.
+
+## Membership updates
+
+Membership updates messages are used to propagate group chat membership changes. The transit format is described in the [Status Payload Specs](status-payload-specs.md). Here we will be describing each specific field.
+
+The format is:
+
+```
+{
+  "events": [],
+  "signature": string,
+  "chat-id": string
+}
+```
+
+### Chat ID
+
+Each membership update MUST be sent with a corresponding `chat-id`. 
+The format of this chat id MUST be a string, [UUID](https://tools.ietf.org/html/rfc4122 ), concatenated with the hex-encoded public key of the creator of the chat. This chat-id MUST be validated by all clients, and should be discarded if it does not follow these rules.
+
+### Signature
+
+The signature for each event is calculated by creating a JSON array of all the `events` sorted by `clock-value` in ascending order, where each event is transformed in an array of tuples `field-name`, `value`. The last element of the array MUST be the `chat-id`. 
+
+For example the event:
+
+```
+   {
+     "chat-id": "chat-id",
+       "events":  [
+       {"b": "b-value"
+        "clock-value": 1,
+        "a": "a-value"
+        },
+       {
+         "e": "e-value",
+         "clock-value": 0,
+         "a": "a-value"
+       }
+       ]
+   }
+
+```
+
+Results in the structure:
+
+```
+   [
+     [
+       [
+         ["a" "a-value"],
+         ["clock-value", 0],
+         ["e" "e-value"]
+       ],
+       [
+         ["a", "a-value"],
+         ["b", "b-value"],
+         ["clock-value", 1]
+       ]
+     ],
+     "chat-id"
+   ]
+```
+
+This structure is then stringified and the `Keccak256` of the string is then signed using its private key by the author and added to the payload.
+      
+
+### Group membership event
+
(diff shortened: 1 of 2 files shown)


status-group-chats-spec.md
```diff
@@ -0,0 +1,202 @@
...
+     "chat-id"
+   ]
+```
+
+This structure is then stringified and the `Keccak256` of the string is then signed using its private key by the author and added to the payload.
+      
+
+### Group membership event
+
+Any group membership event receive MUST be verified using by calculating the signature as in the method described above, and the author MUST be extracted from it, if the verification fails the event MUST be discarded.
```
corpetty: receive[d] spelling.

corpetty: `be verified using by calculating...` seems like it has an extra word in there.

corpetty: `as in the method` -> `as per the method`

corpetty: turn last `,` to `.` and start a new sentence.

status-group-chats-spec.md
```diff
@@ -0,0 +1,202 @@
...
+{
+  "type": "members-added"
+  "members": [string]
+  "clock-value": uint
+}
+```
+
+A members added event is used by admins to add members to the chat.
+Upon receiving this event a client MUST validate the `chat-id` provided with the updates and MUST ensure the author of the event is an admin of the chat, otherwise the event MUST be ignored.
+If the event is valid a client SHOULD update the list of members of the chat who have not joined, adding the `members` received.
```
corpetty: MUST?  otherwise the group is broken, no?

status-group-chats-spec.md
```diff
@@ -0,0 +1,202 @@
...
+{
+  "type": "member-joined"
+  "member": string
+  "clock-value": uint
+}
+```
+
+A members joined event is used by a member of the chat to signal that they want to start receiving messages from this chat.
+Upon receiving this event a client MUST validate the `chat-id` provided with the updates and MUST ensure the author of the event is the same as the one specified by the `member` field.
+If the event is valid a client SHOULD update the list of members of the chat who joined, adding `member`. Any `message` sent to the group chat should now include the newly joined member.
```
corpetty: same as above, otherwise consistency is broken.

status-group-chats-spec.md
```diff
@@ -0,0 +1,202 @@
...
+{
+  "type": "admins-added"
+  "members": [string]
+  "clock-value": uint
+}
+```
+
+An admins added event is used by admins to add make other admins in the chat.
+Upon receiving this event a client MUST validate the `chat-id` provided with the updates, MUST ensure the author of the event is an admin of the chat and MUST ensure all `members` are already `members` of the chat, otherwise the event MUST be ignored.
+If the event is valid a client SHOULD update the list of admins of the chat, adding the `members` received.
```
corpetty: same, MUST for consistency.

status-group-chats-spec.md
```diff
@@ -0,0 +1,202 @@
...
+  "clock-value": uint
+}
+```
+
+A member-removed event is used to leave or kick members of the chat.
+Upon receiving this event a client MUST validate the `chat-id` provided with the updates, MUST ensure that:
+- If the author of the event is an admin, target can only be themselves or a non-admin member.
+- If the author of the event is not an admin, the target of the event can only be themselves.
+-
+If the event is valid a client SHOULD remove the member from the list of `members`/`admins` of the chat, and no further message should be sent to them.
```
corpetty: One more for good measure.

status-group-chats-spec.md
```diff
@@ -0,0 +1,202 @@
...
+  "type": "admin-removed"
+  "member": string
+  "clock-value": uint
+}
+```
+
+An admin-removed event is used to drop admin privileges.
+Upon receiving this event a client MUST validate the `chat-id` provided with the updates, MUST ensure that the author of the event is also the target of the event.
+
+If the event is valid a client SHOULD remove the member from the list of `admins` of the chat.
```
corpetty: just kidding, last one.

status-payloads-spec.md
```diff
@@ -56,43 +67,28 @@ If a signature is not present but an author is provided by a layer below, the me
...
+| 2 | content type | `enum` (more in [Content types](#content-types)) | See details |
+| 3 | message type | `enum` (more in [Message types](#message-types)) | See details |
+| 4 | clock | `int64` | See details |
+| 5 | timestamp | `int64` | See details |
+| 6 | content | `struct { chat-id string, text string, response-to string }` | The chat-id of the chat this message is destined to, the text of the content and optionally the id of the message it is responding to|
 
-### Content types
+#### Content types
 
 Content types are required for a proper interpretation of incoming messages. Not each message is a plain text but may carry a different information.
```
corpetty: can lose the article `a`

status-payloads-spec.md
```diff
@@ -149,10 +141,91 @@ All incoming messages can be matched against a chat. Below you can find a table
 
 <!-- TODO: "group-user-message" is not complete. Does it require to explicitly join the group chat? Is there a way to invite someone? Also, if I start a new group chat (or join an existing one), I need to somehow calculate this chatID by myself. How to do it? -->
 
+### Contact Requests
+
+These messages are used to notify the receiving end that it has been added to the sender's contact. They are identified by the transit tags `c2`, `c3`, `c4` respectively, but they are all interchangeable, meaning a client SHOULD handle them in exactly the same way.
```
corpetty: respective to what?

status-payloads-spec.md
```diff
@@ -149,10 +141,91 @@ All incoming messages can be matched against a chat. Below you can find a table
 
 <!-- TODO: "group-user-message" is not complete. Does it require to explicitly join the group chat? Is there a way to invite someone? Also, if I start a new group chat (or join an existing one), I need to somehow calculate this chatID by myself. How to do it? -->
 
+### Contact Requests
+
+These messages are used to notify the receiving end that it has been added to the sender's contact. They are identified by the transit tags `c2`, `c3`, `c4` respectively, but they are all interchangeable, meaning a client SHOULD handle them in exactly the same way. 
+
+#### Payload
+
+Payload is a struct (a compound data type) with the following fields (order is important):
```
corpetty: not sure the def of struct is necessary

status-payloads-spec.md
```diff
@@ -149,10 +141,91 @@ All incoming messages can be matched against a chat. Below you can find a table
...
+
+
+| Field | Name | Type | Description |
+| ----- | ---- | ---- | ---- |
+| 1 | name | `string` | The self-assigned name of the user (DEPRECATED) |
```
corpetty, line 153: this is an ongoing thing... should we include things that are deprecated in v1?

status-payloads-spec.md
```diff
@@ -149,10 +141,91 @@ All incoming messages can be matched against a chat. Below you can find a table
...
+| 2 | profile image | `string` | The base64 encoded profile picture of the user |
+| 3 | address | `string` | The ethereum address of the user <!-- Why do we need this? can it be computed from the pk? --> |
```
corpetty: no, as v1 separates them.  Multiaccount will continue to use this to define default address in profile.

status-payloads-spec.md
```diff
@@ -149,10 +141,91 @@ All incoming messages can be matched against a chat. Below you can find a table
...
+
+#### Contact update
+
+A client SHOULD send a `ContactUpdate` to all the contacts each time:
+
+- The name is edited
+- The profile image is edited
+- A new device has been paired
+
+A client SHOULD also periodically send a `ContactUpdate` to all the contacts, the interval is up to the client.
```
corpetty: suggest a time period for `periodically`  What do we use?

status-payloads-spec.md
```diff
@@ -149,10 +141,91 @@ All incoming messages can be matched against a chat. Below you can find a table
...
+
+A client SHOULD handle any `Contact*` message in the same way. Any `Contact*` message with a whisper timestamp lower than the last one processed MUST be discarded.
+
+### SyncInstallation
+
+`SyncInstallation` messages are used to synchronize in a best-effort way all the paired installations. It is identified by a transit tag of `p1` 
+
+#### Payload
+
+Payload is a struct (a compound data type) with the following fields (order is important):
```
corpetty: can remove def of struct

status-payloads-spec.md
```diff
@@ -149,10 +141,91 @@ All incoming messages can be matched against a chat. Below you can find a table
...
+| 2 | account | `struct {name string photo-path string last-updated int}` | Information about your own account |
+| 3 | chat | `struct {:public? bool :chat-id string}` | A description of a public chat opened by the client |
+
+### PairInstallation
+
+`PairInstallation` messages are used to propagate informations about a device to its paired devices. It is identified by a transit tag of `p2` 
+
+#### Payload
+
+Payload is a struct (a compound data type) with the following fields (order is important):
```
corpetty: same as previous.

status-payloads-spec.md
```diff
@@ -149,10 +141,91 @@ All incoming messages can be matched against a chat. Below you can find a table
...
+| 4 | fcm-token | `string` | The FCM Token used by mobile devices for push notifications |
+
+### GroupMembershipUpdate
+
+`GroupMembershipUpdate` is a message used to propagate information about group membership changes in a group chat.. It is identified by a transit tag of `g5`.
+The details are in the  [Group chats specs](status-group-chats-spec.md)
+
+#### Payload
+
+Payload is a struct (a compound data type) with the following fields (order is important):
```
corpetty: same as above

status-payloads-spec.md
```diff
@@ -149,10 +141,91 @@ All incoming messages can be matched against a chat. Below you can find a table
...
+| 2 | membership-updates | See details | A list of events that describe the membership changes |
+| 3 | message | `Transit message` | An optional message, described in [Message](#message) |
+
 ## Upgradability
 
-The current protocol format is hardly upgradable without breaking backward compatibility. Because Transit is used in this particular way described above, the only reliable option is to append a new field to the Transit record definition. It will be simply ignored by the old clients.
-<!-- Not sure I agree with this statement, seems very arbitrary, appending to an array is just as upgradable as adding an entry in a map, just less convenient, I would remove the qualitative statement, and just describe how to upgrade -->
+There are two ways to upgrade the protocol without breaking compatibility:
+
+- Map fields can be enriched with a new key, which will be ignored by old clients.
```
corpetty: what are map fields?

status-payloads-spec.md
```diff
@@ -162,5 +235,4 @@ TBD.
 
 ### Why are you using Transit and Protobuf?
 
-Transit was initially chose for encoding, and Protobuf was added afterwards. This is partly due to the history of the protocol living inside of `status-react`, which is written in Clojurescript. In future versions of payload and data sync client specifications it is likely we'll move towards Protobuf only. See e.g. [Dasy](https://github.com/vacp2p/dasy) for a research proof of concept.
-<!-- I would remove the link to dasy, I find it a bit confusing, and the repo just implements something totally different and a fraction of the functionalities -->
+Transit was initially chose for encoding, and Protobuf was added afterwards. This is partly due to the history of the protocol living inside of `status-react`, which is written in Clojurescript.
```
corpetty, line 227: do we have consensus on the future of this?  Are we planning on removing transit?

corpetty (changes requested):

status-group-chats-spec.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Membership updates
+
+Membership updates messages are used to propagate group chat membership changes. The transit format is described in the [Status Payload Specs](status-payload-specs.md). Here we will be describing each specific field.
+
+The format is:
+
+```
+{
+  "events": [],
```
adambabik: It is a slice of what types?

status-group-chats-spec.md
```diff
@@ -0,0 +1,202 @@
...
+  "signature": string,
+  "chat-id": string
+}
+```
+
+### Chat ID
+
+Each membership update MUST be sent with a corresponding `chat-id`. 
+The format of this chat id MUST be a string, [UUID](https://tools.ietf.org/html/rfc4122 ), concatenated with the hex-encoded public key of the creator of the chat. This chat-id MUST be validated by all clients, and should be discarded if it does not follow these rules.
```
adambabik:
```suggestion
The format of this chat id MUST be a string, [UUID](https://tools.ietf.org/html/rfc4122 ), concatenated with the hex-encoded public key of the creator of the chat. This chat-id MUST be validated by all clients, and SHOULD be discarded if it does not follow these rules.
```

status-group-chats-spec.md
```diff
@@ -0,0 +1,202 @@
...
+         "a": "a-value"
+       }
+       ]
+   }
+
+```
+
+Results in the structure:
+
+```
```
adambabik, line 75: What about the order of fields? It's undefined in JSON objects. If we transform it in this way, we need to impose some order so that all clients do it in the same way.

status-group-chats-spec.md
```diff
@@ -0,0 +1,202 @@
...
+         ["a", "a-value"],
+         ["b", "b-value"],
+         ["clock-value", 1]
+       ]
+     ],
+     "chat-id"
+   ]
+```
+
+This structure is then stringified and the `Keccak256` of the string is then signed using its private key by the author and added to the payload.
```
adambabik: `stringified` is too board. What about white spaces?

status-payloads-spec.md
```diff
@@ -149,10 +141,91 @@ All incoming messages can be matched against a chat. Below you can find a table
...
+| ----- | ---- | ---- | ---- |
+| 1 | name | `string` | The self-assigned name of the user (DEPRECATED) |
+| 2 | profile image | `string` | The base64 encoded profile picture of the user |
+| 3 | address | `string` | The ethereum address of the user <!-- Why do we need this? can it be computed from the pk? --> |
+| 4 | fcm-token | `string` | The FCM Token used by mobile devices for push notifications (DEPRECATED) |
+| 5 | device-info | `[struct { id string, fcm-token string }]` | A list of pair `installation-id`, `fcm-token` for each device that is currently paired |
+
+#### Contact update
+
+A client SHOULD send a `ContactUpdate` to all the contacts each time:
```
adambabik, line 161: Is payload the same as described above?

status-payloads-spec.md
```diff
@@ -149,10 +141,91 @@ All incoming messages can be matched against a chat. Below you can find a table
...
+- The name is edited
+- The profile image is edited
+- A new device has been paired
+
+A client SHOULD also periodically send a `ContactUpdate` to all the contacts, the interval is up to the client.
+
+
+#### Handling contact messages
+
+A client SHOULD handle any `Contact*` message in the same way. Any `Contact*` message with a whisper timestamp lower than the last one processed MUST be discarded.
```
adambabik, line 172: I noticed we try to remove Whisper dependency from this spec but I believe not really possible here. There is no other field that can replace `whisper timestamp` to validate which contact update should be discarded, right?

status-payloads-spec.md
cammellos: fields that are hashmaps/associative arrays/structs i.e {"something-like-this": "a"} maybe best to call it a struct for consistency?

status-payloads-spec.md
cammellos: We have consensus that we want to remove transit, but given that if that happens needs to happen before code freeze, not sure we will get around, so I would assume that it does not happen unless it does :)

status-group-chats-spec.md
cammellos: we sort them alphabetically ascending, forgot to add that to the description

status-payloads-spec.md
cammellos: Currently no, we could use a logical clock (clock-value), but not sure we will get around before code freeze

status-payloads-spec.md
cammellos: Edited at last moment and left it out, added now

status-payloads-spec.md
cammellos: I think until we get around actually making sure that we can drop them yes, basically as I understand this field is not used anymore, but if you drop it a client might not accept the message as it fails validation, so unless we get around fixing for code freeze unfortunately we need to keep it as it is

status-payloads-spec.md
cammellos: 48 hours currently, updated

status-payloads-spec.md
cammellos: Yes, `ContactUpdate`/`ContactRequest`/`ContactRequestConfirmed` have exactly the same payload, but different transit tag (we should probably just have a single message)

cammellos:
@corpetty @adambabik addressed the feedback, thanks. ready to be reviewed again.

status-payloads-spec.md
adambabik: Should we make it more explicit?

cammellos: done!

adambabik (approved):

status-payloads-spec.md
corpetty: I do like consistency, but if the naming is specifically appropriate then we can use it, as there are some subtle differences to those.

status-payloads-spec.md
corpetty: keep them both, just curious.  I'm definitely on the plan of mapping out exactly how v1 works, then we can discuss how to move forward.

corpetty (approved):

PR #60, opened 2019-11-12
Describe rate limiting in Whisper

### LINKED ISSUE #66: Describe the rate limiting feedback for peers
## Problem

Due to using a low PoW, [Status nodes and clients are vulnerable](https://github.com/status-im/trailofbits-audit/issues/35) to a denial of service attack.

It's not possible to increase PoW because of mobile usage. The alternative is an [implementation of rate limiting](https://github.com/status-im/whisper/pull/33). 

## Solution

Describe in the spec, how rate limiting would work. It should take into consideration:
- the general algorithm 
- rate limiting settings on the client and server
- how rate limiting information is broadcasted to peers (similar to broadcasting PoW and bloom filters settings)
- rate limiting information in the handshake.

adambabik:
Closes #66

### DIFF
--- a/status-whisper-usage-spec.md
+++ b/status-whisper-usage-spec.md
@@ -62,11 +62,25 @@ Whisper's Proof Of Work algorithm is used to deter denial of service and various
 
 ## Rate limiting
 
-In order to provide a very basic Denial-of-Service attack protection, there SHOULD be a rate limiting applied to leaf Whisper nodes, i.e. nodes not being controler within a cluster.
+In order to provide an optional very basic Denial-of-Service attack protection, each node SHOULD define its own rate limits. The rate limits SHOULD be applied on IPs, peer IDs, and envelope topics.
 
-Currently, the rate limiting is applied on peer IDs and IPs from which the packets come from. It is still in progress to decide on specific numbers. The Whisper nodes within a cluster SHOULD be whitelisted to avoid dropping connection when broadcasting messages between each other. When a limit is hit, the packet or the connection MUST be dropped (TBD).
+Each node MAY decide to whitelist, i.e. do not rate limit, selected IPs or peer IDs.
 
-The rate limiting SHOULD be also applied on a topic level in order to prevent a peer to spam a particular channel. The specific limit per peer per topic is to be decided.
+If a peer exceeds node's rate limits, the connection between them MAY be dropped.
+
+Each node SHOULD broadcast its rate limits to its peers using rate limits packet code (`0x14`). The rate limits is RLP-encoded information:
+
+```
+[ IP limits, PeerID limits, Topic limits ]
+```
+
+`IP limits`: 4-byte wide unsigned integer
+`PeerID limits`: 4-byte wide unsigned integer
+`Topic limits`: 4-byte wide unsigned integer
+
+The rate limits MAY be also sent as an optional parameter in the handshake.
+
+Each node SHOULD record rate limits advertised by its peers and adjust the number of sent packets in order not to exceed peer's rate limits. If the limit gets exceeded, the connection MAY be dropped.
 
 <!-- TODO: provide an instruction how to start a Whisper node with proper configuration using geth.-->
 
--- a/status-whisper-usage-spec.md
+++ b/status-whisper-usage-spec.md
@@ -9,12 +9,12 @@
   - [Reason](#reason)
   - [Terminology](#terminology)
   - [Whisper node configuration](#whisper-node-configuration)
+  - [Handshake](#handshake)
   - [Rate limiting](#rate-limiting)
   - [Keys management](#keys-management)
     - [Contact code topic](#contact-code-topic)
     - [Partitioned topic](#partitioned-topic)
     - [Public chats](#public-chats)
-    - [Personal discovery topic](#personal-discovery-topic)
     - [Generic discovery topic](#generic-discovery-topic)
     - [One-to-one topic](#one-to-one-topic)
     - [Group chat topic](#group-chat-topic)
@@ -60,6 +60,20 @@ Whisper's Proof Of Work algorithm is used to deter denial of service and various
 * proof-of-work requirement not larger than `0.002`
 * time-to-live not lower than `10` (in seconds)
 
+## Handshake
+
+Handshake is a RLP-encoded packet sent to a newly connected peer. It MUST start with a Status Code (`0x00`) and follow up with items:
+```
+[ protocolVersion, PoW, bloom, isLightNode, confirmationsEnabled, rateLimits ]
+```
+
+`protocolVersion`: version of the Whisper protocol
+`PoW`: minimum PoW accepted by the peer
+`bloom`: bloom filter of Whisper topic accepted by the peer
+`isLightNode`: when true, the peer won't forward messages
+`confirmationsEnabled`: when true, the peer will send message confirmations
+`rateLimits`: is `[ RateLimitIP, RateLimitPeerID, RateLimitTopic ]` where each values is an integer with a number of accepted packets per second per IP, Peer ID, and Topic respectively
+
 ## Rate limiting
 
 In order to provide an optional very basic Denial-of-Service attack protection, each node SHOULD define its own rate limits. The rate limits SHOULD be applied on IPs, peer IDs, and envelope topics.
@@ -80,11 +94,7 @@ Each node SHOULD broadcast its rate limits to its peers using rate limits packet
 
 The rate limits MAY also be sent as an optional parameter in the handshake.
 
-Each node SHOULD record rate limits advertised by its peers and adjust the number of sent packets in order not to exceed peer's rate limits. If the limit gets exceeded, the connection MAY be dropped.
-
-<!-- TODO: provide an instruction how to start a Whisper node with proper configuration using geth.-->
-
-<!-- @TODO: is there a higher bound -->
+Each node SHOULD respect rate limits advertised by its peers. The number of packets SHOULD be throttled in order not to exceed peer's rate limits. If the limit gets exceeded, the connection MAY be dropped by the peer.
 
 ## Keys management
 
--- a/status-whisper-usage-spec.md
+++ b/status-whisper-usage-spec.md
@@ -9,6 +9,7 @@
   - [Reason](#reason)
   - [Terminology](#terminology)
   - [Whisper node configuration](#whisper-node-configuration)
+  - [Rate limiting](#rate-limiting)
   - [Keys management](#keys-management)
     - [Contact code topic](#contact-code-topic)
     - [Partitioned topic](#partitioned-topic)
@@ -20,7 +21,7 @@
   - [Message encryption](#message-encryption)
   - [Whisper V6 extensions](#whisper-v6-extensions)
     - [Request historic messages](#request-historic-messages)
-      - [shhext_requestMessages](#shhext_requestmessages)
+      - [shhext_requestMessages](#shhextrequestmessages)
 
 ## Abstract
 
@@ -58,6 +59,14 @@ If you want to run a Whisper node and receive messages from Status clients, it m
 Whisper's Proof Of Work algorithm is used to deter denial of service and various spam/flood attacks against the Whisper network. The sender of a message must perform some work which in this case means processing time. Because Status' main client is a mobile client, this easily leads to battery draining and poor performance of the app itself. Hence, all clients MUST use the following Whisper node settings:
 * proof-of-work requirement not larger than `0.002`
 * time-to-live not lower than `10` (in seconds)
+
+## Rate limiting
+
+In order to provide a very basic Denial-of-Service attack protection, there SHOULD be a rate limiting applied to leaf Whisper nodes, i.e. nodes not being controler within a cluster.
+
+Currently, the rate limiting is applied on peer IDs and IPs from which the packets come from. It is still in progress to decide on specific numbers. The Whisper nodes within a cluster SHOULD be whitelisted to avoid dropping connection when broadcasting messages between each other. When a limit is hit, the packet or the connection MUST be dropped (TBD).
+
+The rate limiting SHOULD be also applied on a topic level in order to prevent a peer to spam a particular channel. The specific limit per peer per topic is to be decided.
 
 <!-- TODO: provide an instruction how to start a Whisper node with proper configuration using geth.-->
 
@@ -145,7 +154,7 @@ for i = 0; i < topicLen; i++ {
 }
 ```
 
-<!-- NOTE: commented out as it is currently not used.  In code for potential future use. - C.P. Oct 8, 2019 
+<!-- NOTE: commented out as it is currently not used.  In code for potential future use. - C.P. Oct 8, 2019
 ### Personal discovery topic
 
  Personal discovery topic is used to ???
@@ -169,7 +178,7 @@ for i = 0; i < topicLen; i++ {
 
 Each Status Client SHOULD listen to this topic in order to receive ??? -->
 
-<!-- NOTE: commented out as it is no longer valid as of V1. - C.P. Oct 8, 2019 
+<!-- NOTE: commented out as it is no longer valid as of V1. - C.P. Oct 8, 2019
 ### Generic discovery topic
 
 Generic discovery topic is a legacy topic used to handle all one-to-one chats. The newer implementation should rely on [Partitioned Topic](#partitioned-topic) and [Personal discovery topic](#personal-discovery-topic).
--- a/status-whisper-usage-spec.md
+++ b/status-whisper-usage-spec.md
@@ -78,7 +78,7 @@ Each node SHOULD broadcast its rate limits to its peers using rate limits packet
 `PeerID limits`: 4-byte wide unsigned integer
 `Topic limits`: 4-byte wide unsigned integer
 
-The rate limits MAY be also sent as an optional parameter in the handshake.
+The rate limits MAY also be sent as an optional parameter in the handshake.
 
 Each node SHOULD record rate limits advertised by its peers and adjust the number of sent packets in order not to exceed peer's rate limits. If the limit gets exceeded, the connection MAY be dropped.
 
--- a/status-whisper-usage-spec.md
+++ b/status-whisper-usage-spec.md
@@ -74,6 +74,7 @@ Handshake is a RLP-encoded packet sent to a newly connected peer. It MUST start 
 `confirmationsEnabled`: when true, the peer will send message confirmations
 `rateLimits`: is `[ RateLimitIP, RateLimitPeerID, RateLimitTopic ]` where each values is an integer with a number of accepted packets per second per IP, Peer ID, and Topic respectively
 
+`bloom, isLightNode, confirmationsEnabled, and rateLimits` are all optional arguments in the handshake. However, if you specify optional field you MUST also specify all optional fields preceding it, in order to be unambiguous.
 ## Rate limiting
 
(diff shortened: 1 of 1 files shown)


status-whisper-usage-spec.md
```diff
@@ -59,6 +60,14 @@ Whisper's Proof Of Work algorithm is used to deter denial of service and various
 * proof-of-work requirement not larger than `0.002`
 * time-to-live not lower than `10` (in seconds)
 
+## Rate limiting
+
+In order to provide a very basic Denial-of-Service attack protection, there SHOULD be a rate limiting applied to leaf Whisper nodes, i.e. nodes not being controler within a cluster.
```
oskarth: what's the 'leaf Whisper node' distinction about? strikes me as odd, a protocol shouldn't take something like "a cluster" as for granted IMO

controller/controlled typo

status-whisper-usage-spec.md
```diff
@@ -59,6 +60,14 @@ Whisper's Proof Of Work algorithm is used to deter denial of service and various
...
+
+Currently, the rate limiting is applied on peer IDs and IPs from which the packets come from. It is still in progress to decide on specific numbers. The Whisper nodes within a cluster SHOULD be whitelisted to avoid dropping connection when broadcasting messages between each other. When a limit is hit, the packet or the connection MUST be dropped (TBD).
```
oskarth: Not sure about the use of "Currently" "in the future/TBD" - for temporal WIP stuff lets put in comment and only describe _what is_ in the spec

status-whisper-usage-spec.md
```diff
@@ -59,6 +60,14 @@ Whisper's Proof Of Work algorithm is used to deter denial of service and various
...
+
+The rate limiting SHOULD be also applied on a topic level in order to prevent a peer to spam a particular channel. The specific limit per peer per topic is to be decided.
```
oskarth: "to be decided" isn't spec language

oskarth:
Thanks for PR! I know it is just a draft, but I made some comments already

status-whisper-usage-spec.md
adambabik: That's why it is a draft PR. I haven't finalized the numbers. When I do, we can be more specific. However, I don't even know whether we should put any specific numbers... It depends on a situation.

status-whisper-usage-spec.md
kdeme: I guess leaf node is the wrong wording here? 
Are there actual leaf nodes right now? Doesn't each node currently have multiple peers (although perhaps always other cluster nodes?).

So it is rather "whitelisted cluster nodes" versus "the rest". I don't think I like that distinction.

kdeme (commented):
From reading the spec, I'm assuming that rate limiting is applied only on the incoming messages (haven't checked the code). The general problem I see with this type of rate limiting is that it seems to work against the decentralized idea where anyone can start a node, due to the broadcasting nature of the Whisper protocol.

I assume that the `leaf node` distinction is made so that a `cluster` node does not get disconnected because e.g. two of its peers send together more than the incoming rated limit per peer (but less individually), within the interval of sending the queue, which in its turn can cause disconnects for this cluster node when forwarding those messages to his other peers.

The issue I see here is that having this distinction does not allow for anyone to spin up there own node for the network (even though there is no incentive to do this now, we should not start disconnecting them either because they are not part of the "cluster"...). Or am I misunderstanding on how this would work?

status-whisper-usage-spec.md
kdeme: I guess leaf node is the wrong wording here?
Are there actual leaf nodes right now? Doesn't each node currently have multiple peers (although perhaps always other cluster nodes?).

So it is rather "whitelisted cluster nodes" versus "the rest". I don't think I like that distinction.

decanus:
```suggestion
In order to provide a very basic Denial-of-Service attack protection, rate limiting SHOULD be applied to leaf Whisper nodes, i.e. nodes not being controller within a cluster.
```

status-whisper-usage-spec.md
decanus:
```suggestion
The rate limiting SHOULD also be applied on the topic level in order to prevent a peer from spamming a particular channel. The specific limit per peer per topic is to be decided.
```

decanus (changes requested):

status-whisper-usage-spec.md
adambabik: Yeah, the current Status client nodes are are leaf nodes but this term is not used anywhere. The other is light nodes so the nodes which do not broadcast messages.

> So it is rather "whitelisted cluster nodes" versus "the rest". I don't think I like that distinction.

This is the most accurate. Not sure we can do it other way around. This is the reality where we have nodes which produce messages (and we want to rate limit them) and nodes which pass them along and we don't want limit these.

adambabik:
> From reading the spec, I'm assuming that rate limiting is applied only on the incoming messages (haven't checked the code). The general problem I see with this type of rate limiting is that it seems to work against the decentralized idea where anyone can start a node, due to the broadcasting nature of the Whisper protocol.

Of course, this solution is local to a cluster. The "proper" solution is PoW but we can't apply it on mobile clients.

> The issue I see here is that having this distinction does not allow for anyone to spin up there own node for the network (even though there is no incentive to do this now, we should not start disconnecting them either because they are not part of the "cluster"...). Or am I misunderstanding on how this would work?

Correct. Such a new node would need to be whitelisted. It's a kind of a problem we currently have. There might be dishonest peers in the current network. The client selects only 1-2 Whisper nodes. If it selects only dishonest peers, it will receive only filtered messages. The solution we were thinking about long time ago was a smart contact where honest mailservers could be collected. They would need to stake some SNT + validators run by Status (in the simplest form) would need to exist. In the case of whitelisted peers, we would need a similar system I believe.

Generally, we talked about it in an issue and this is the most naive but working solution in the current reality. Unless there is some better idea...

oskarth:
I don't understand why we need to distinguish between cluster nodes and normal nodes to apply rate limiting. Why can't it be universally applied without any kind of special treatment? Enshrining our cluster as special is a huge issue.

adambabik:
@oskarth ok, so what's your proposal? How do you want to rate limit communication between server node A and server node B where the first one has, e.g. 100 peers? It means that each of 100 peers can produce 1 message per second but node A would send 100 message per second to node B. That's why Whisper implemented PoW instead of rate limiting because it's impossible to have global rate limiting allowing all scenarios.

Instead of whitelisting we can rate limit only light nodes but then it's enough for an attacker to run a node on the server and turn off light node mode and DoS all our clients.

Alternatively, we can apply rate limiting only on the Status clients (and mailservers likely). In such a case, it's still possible to spam the network but at least do not kill the clients (clients would disconnect from spamming peers). But I am not sure this fulfils ToB requirements. cc @corpetty

oskarth:
@adambabik Rate limit by default for all nodes. Provide whitelisting as an _option_ for all nodes.

Cluster whitelist their own nodes. That's an implementation details. Keep specialized trust / deployment specifics at that level, not protocol level.

status-whisper-usage-spec.md
corpetty: I agree that this should be a generic setting on the node, which is tweaked based on the context.

status-whisper-usage-spec.md
corpetty: Do we have any information on the percentage of redundant envelops received from peers?

oskarth:
@adambabik will you update this based on decision in Monday's call? Also please add relevant issue to Waku project and replace existing placeholder.

> Rate limit: Packet code feedback for rate limiting [needs more detail] - Adam

adambabik:
@oskarth yes, let's confirm the decision:
1. Rate limiting is a setting per node; each node needs to enable it and set the rate limits,
2. Whitelisting peers and IPs is possible,
3. A node broadcasts its rate limits (that would need to be in a handshake as well),
4. In the case of a violation, a peer might be disconnected similarly to sending a message with invalid PoW etc.

What is specific to our server nodes, we will white list all our servers and decide on the rate limits. The client (Status app) will have different numbers and also will whitelist known peers.

adambabik:
@oskarth @decanus please review again.

status-whisper-usage-spec.md
```diff
@@ -62,11 +62,25 @@ Whisper's Proof Of Work algorithm is used to deter denial of service and various
...
+
+```
+[ IP limits, PeerID limits, Topic limits ]
+```
+
+`IP limits`: 4-byte wide unsigned integer
+`PeerID limits`: 4-byte wide unsigned integer
+`Topic limits`: 4-byte wide unsigned integer
+
+The rate limits MAY be also sent as an optional parameter in the handshake.
```
oskarth: Does it have to be? Vs just its own Status code, and communicating right after handshake

adambabik: The best would be to do it in the handshake, otherwise, a node needs to restrain any activity before the rate limits are broadcasted and peers handle it which is unknown when it exactly happens.

decanus:
once this is merged we should add it to Waku, or should it already be added @oskarth

oskarth:
@decanus I think we can wait until this is merged, and then add it to Waku along with implementation matrix update.

@adambabik @kdeme any more thoughs on the handshake? It seems a bit iffy now with existing layout https://github.com/vacp2p/specs/pull/52#discussion_r351096293 previous clients and now this change. Can we simplify things somehow or communicate things in a more open-ended vs positional way? How can we ensure better consistency here? Both for existing spec, this addition, and going forward (moar capabilities).

status-whisper-usage-spec.md
```diff
@@ -59,6 +60,28 @@ Whisper's Proof Of Work algorithm is used to deter denial of service and various
...
+
+```
+[ IP limits, PeerID limits, Topic limits ]
+```
+
+`IP limits`: 4-byte wide unsigned integer
+`PeerID limits`: 4-byte wide unsigned integer
+`Topic limits`: 4-byte wide unsigned integer
+
+The rate limits MAY also be sent as an optional parameter in the handshake.
```
oskarth, line 96: This needs more detail, how exactly will the handshake be added optionally? See comment below

status-whisper-usage-spec.md
```diff
@@ -59,6 +60,28 @@ Whisper's Proof Of Work algorithm is used to deter denial of service and various
...
 
+## Rate limiting
+
+In order to provide an optional very basic Denial-of-Service attack protection, each node SHOULD define its own rate limits. The rate limits SHOULD be applied on IPs, peer IDs, and envelope topics.
+
+Each node MAY decide to whitelist, i.e. do not rate limit, selected IPs or peer IDs.
+
+If a peer exceeds node's rate limits, the connection between them MAY be dropped.
+
+Each node SHOULD broadcast its rate limits to its peers using rate limits packet code (`0x14`). The rate limits is RLP-encoded information:
```
oskarth, line 86: Why 0x14?

oskarth (changes requested):
Overall looks good to me, once we figure out handshake this seems fine to me.

@corpetty what about ToB?

oskarth:
One thing we might do is append to handshake for Whisper v6, and then clarify in spec that only unambiguous patterns are legal. Example, with a,b,c being optional fields:

`required`
-`required a`
- `required a b`
- `required a b c`

This is illegal:
- `required b`
- `...c`, `... a c`, ... `b c`

Then, for waku v1, we ensure we have an extensible design that allows unambiguous optionals.

What do people think of this approach?

status-whisper-usage-spec.md
adambabik: We can either start from some further number to allow more message related packet codes to be in front or just continue from the last used.

adambabik:
This is the current situation:

EIP-627 (based on geth implementation):
```
(statusCode, protocolVersion[, pow][, bloom][, isLightNode])
```

It means that `pow`, `bloom` and `isLightNode` are optional. However, it's possible to skip `pow` and `bloom` and pass only `isLightNode`. This is odd.

Our Whisper fork:
```
(statusCode, protocolVersion[, pow][, bloom][, isLightNode][, confirmationsEnabled])
```
It means that `pow`, `bloom`, `isLightNode` and `confirmationsEnabled` are optional. However, it's possible to skip `pow` and `bloom` and pass only `isLightNode` or only `confirmationsEnabled`. Situation where both `isLightNode` and `confirmationsEnabled` are passed but not `pow` and `bloom` is also allowed.

As you can see it's very complicated and unclear. Because it's our protocol spec, I would go for:
```
(statusCode, protocolVersion[, pow][, bloom][, isLightNode][, confirmationsEnabled][, rateLimits])
```
where `rateLimits` is optional and can be specified independently from the rest of the optional parameters.

In Waku/1, I would go for a scheme where anything after `protocolVersion` is optional but if one wants to pass, for example, `isLightNode` then all previous arguments need to be passed as well.

kdeme:
So according to EIP-627, `PoW` is not optional. Just bloom filter is optional. `isLightNode` is not in spec but is in geth / nim-eth.

For nim-eth these optional arguments will only work in one direction for now: see https://github.com/status-im/nim-eth/issues/116 . But we will need to fix this it seems (As we have  `confirmationsEnabled` added. And now there is also rate limits and probably in the future Waku mode / topics).
I'd have to further think about it but the last point of @adambabik might make sense to make it more easy. Most simple solution would be no "optionally" at all :)

kdeme:
Regarding the rate limiting specification:

It looks like an improvement on the current "too low" PoW situation. Seems like it could stop a DoS from spreading over the network, or at least a cluster within the network, and still allow individual nodes.

Perhaps this is obvious, but I would like to note that it will still be possible (D)DoS a single node, as on the output node messages will get dropped unrelated to where the traffic comes from. E.g. One peer could stay under the limits with one connected node. But the node might go over the limits with multiple peers. But OK, this can get solved with whitelisting & somewhat with good configuration (or at least make it more difficult -> more IPs).

status-whisper-usage-spec.md
```diff
@@ -59,6 +60,28 @@ Whisper's Proof Of Work algorithm is used to deter denial of service and various
...
+[ IP limits, PeerID limits, Topic limits ]
+```
+
+`IP limits`: 4-byte wide unsigned integer
+`PeerID limits`: 4-byte wide unsigned integer
+`Topic limits`: 4-byte wide unsigned integer
+
+The rate limits MAY also be sent as an optional parameter in the handshake.
+
+Each node SHOULD record rate limits advertised by its peers and adjust the number of sent packets in order not to exceed peer's rate limits. If the limit gets exceeded, the connection MAY be dropped.
```
decanus: I'm not 100% sure what this means

adambabik:
@kdeme 
> So according to EIP-627, PoW is not optional. Just bloom filter is optional. isLightNode is not in spec but is in geth / nim-eth.

Where is it stated? It is optional in the implementation, just saying.

> E.g. One peer could stay under the limits with one connected node. But the node might go over the limits with multiple peers.

Yeah, that's possible but rate limiting is defined per peer so if a node wants to receive max 10 messages per second and it is willing to connect to 10 peers, it may advertise the rate limiting of 1 message per second. All peers should respect that and throttle the number of sent messages. This may lead to missing some messages though.

status-whisper-usage-spec.md
adambabik: Assume Node A having Peer1, Peer2 and Pee3. Node A advertised rate limits of 5 messages per second per peer. Peer1 should throttle broadcasting messages to Node A and respect the limit of 5 messages per second. This brings a dangerous problem of skipping some messages.

decanus: Can this be better formulated in the spec, not sure how yet.

oskarth:
>  Document further differences with Whisper v6
@kdeme any chance you could take a look at this one? Changelog in spec has section "Differences between shh/6 waku/0" and it'd be great if all the major differences were there. E.g. handshake business and whatnot

oskarth:
> >  So according to EIP-627, PoW is not optional. Just bloom filter is optional. isLightNode is not in spec but is in geth / nim-eth.
>
> Where is it stated? It is optional in the implementation, just saying.

From https://eips.ethereum.org/EIPS/eip-627:

> This packet contains two objects: integer message code (0x00) followed by a list of values: integer version, float PoW requirement, and bloom filter, in this order. The bloom filter paramenter is optional; if it is missing or nil, the node is considered to be full node (i.e. accepts all messages). The format of PoW and bloom filter please see below (message codes 2 and 3).

It only mentions bloom filter as optional, which implies the previous fields are required.

---

I'm not so sure about having `required optional-a optional-b` where you can omit `optional-a` and specify `optional-b`, seems very fragile. What's the issue with locking down as I proposed above?

status-whisper-usage-spec.md
oskarth: since whisper is queue based, maybe talking in those terms would be useful

adambabik: > Each node SHOULD respect rate limits advertised by its peers. The number of packets SHOULD be throttled in order not to exceed peer's rate limits. If the limit gets exceeded, the connection MAY be dropped by the peer.

Better?

adambabik:
Thanks @oskarth for that fragment. Yeah, so the implementation does not follow the spec...

> I'm not so sure about having required optional-a optional-b where you can omit optional-a and specify optional-b, seems very fragile. What's the issue with locking down as I proposed above?

We either write this spec as it is or write it as it should be. The current implementation allows to omit some `optional-a` parameters and specify `optional-b`. My proposal for Waku/1 is more strict and we can lock it down this way but we would need to change the implementation and won't be compatible with EIP-627. Given that the current way is not really viable in Nim maybe we should go with the more strict one.

kdeme:
I also prefer the more strict approach as suggested, it is likely to make implementation less complex.

I don't think the strict approach collides with EIP-627. The fragment @oskarth posted also mentions `in this order`. In my interpretation, when you omit a field you automatically lose order.

adambabik:
Ok, so I think the consensus is to make the spec strict even though the current state of things in Go implementation is a bit different. In my opinion, it's a good decision without negative consequences.  I will update this PR. 

Thanks all for contributing to the discussion!

status-whisper-usage-spec.md
```diff
@@ -59,9 +60,41 @@ Whisper's Proof Of Work algorithm is used to deter denial of service and various
...
+[ protocolVersion, PoW, bloom, isLightNode, confirmationsEnabled, rateLimits ]
+```
+
+`protocolVersion`: version of the Whisper protocol
+`PoW`: minimum PoW accepted by the peer
+`bloom`: bloom filter of Whisper topic accepted by the peer
+`isLightNode`: when true, the peer won't forward messages
+`confirmationsEnabled`: when true, the peer will send message confirmations
+`rateLimits`: is `[ RateLimitIP, RateLimitPeerID, RateLimitTopic ]` where each values is an integer with a number of accepted packets per second per IP, Peer ID, and Topic respectively
+
```
oskarth, line 76:
```suggestion

`bloom, isLightNode, confirmationsEnabled, and rateLimits` are all optional arguments in the handshake. However, if you specify optional field you MUST also specify all optional fields preceding it, in order to be unambiguous.
```

oskarth: This was my initial suggestion and it is stricter but also maintains compatibility with EIP 627. @kdeme @adambabik do you see any issues with this? Or would you definitely prefer an even stricter version, where all fields are always required (thus breaking compatibility with EIP-627)?

oskarth: (Right now it is ambiguous because current EIP627 makes bloom filter optional, and this spec diff also mentions "optional rate limiting". Also typo in including 'and' in above suggestion quoted form)

adambabik: I don't see any issues. I don't think there are any clients that specify some optional field without others.

oskarth (approved):

PR #64, opened 2019-11-21
Fix mailserver spec

### LINKED ISSUE #55: Fix Whisper usage spec wrt mailserver
As a client implementer, I want specs to describe how to use mailservers, not how the Status app works.

## Right now
Currently this spec: https://github.com/status-im/specs/blob/master/status-whisper-usage-spec.md#request-historic-messages explains the boring / non protocoly part: status-react to status go.

## Expected

Instead, it should explain the interface between status-go and mailservers. Specifically:

"The above topics is then converted into a bloom filter and then and sent to the Mailserver." should be expanded on, and the "Object - The message request object:  [...]" part from status-react should probably be removed.

^ @adambabik @corpetty

adambabik:
Closes #55

### DIFF
--- a/status-whisper-mailserver-spec.md
+++ b/status-whisper-mailserver-spec.md
@@ -29,13 +29,13 @@ It is important to notice that Mailserver will only handle requests from its dir
 
 ### Archiving messages
 
-A node which wants to provide a mailserver capability MUST store envelopes incoming in messages packets identify by a code `0x01`. The envelopes can be stored in an arbitrary format, however, they MUST be serialized and deserialized to Whisper envelope format.
+A node which wants to provide mailserver functionality MUST store envelopes from incoming messages packets (Whisper packet-code `0x01`). The envelopes can be stored in an arbitrary format, however, they MUST be serialized and deserialized to Whisper envelope format.
 
-A mailserver SHOULD store envelopes for all topics to be useful by any peer, however, when for personal use case it MAY store only subset of topics.
+A mailserver SHOULD store envelopes for all topics to be generally useful for any peer, however, for specific use cases it MAY store envelopes for a subset of topics.
 
 ### Requesting messages
 
-In order to request historic messages, a node MUST send a packet P2P Request (`0x7e`) to its peer providing mailserver capability. This packet requires one argument which MUST be a Whisper envelope.
+In order to request historic messages, a node MUST send a packet P2P Request (`0x7e`) to its peer providing mailserver functionality. This packet requires one argument which MUST be a Whisper envelope.
 
 In the Whisper envelope's payload section, there MUST be RLP-encoded information about the details of the request:
 
--- a/status-whisper-mailserver-spec.md
+++ b/status-whisper-mailserver-spec.md
@@ -35,7 +35,7 @@ A mailserver SHOULD store envelopes for all topics to be generally useful for an
 
 ### Requesting messages
 
-In order to request historic messages, a node MUST send a packet P2P Request (`0x7e`) to its peer providing mailserver functionality. This packet requires one argument which MUST be a Whisper envelope.
+In order to request historic messages, a node MUST send a packet P2P Request (`0x7e`) to a peer providing mailserver functionality. This packet requires one argument which MUST be a Whisper envelope.
 
 In the Whisper envelope's payload section, there MUST be RLP-encoded information about the details of the request:
 
@@ -49,13 +49,15 @@ In the Whisper envelope's payload section, there MUST be RLP-encoded information
 `Limit`: 4-byte wide unsigned integer limiting the number of returned envelopes
 `Cursor`: 32-byte wide array of a cursor returned from the previous request (optional)
 
+The `Cursor` field SHOULD be filled in if a number of envelopes between `Lower` and `Upper` is greater than `Limit` so that the requester can send another request using the obtained `Cursor` value. What exactly is in the `Cursor` is up to the implementation. The requester SHOULD NOT use a `Cursor` obtained from one mailserver in a request to another mailserver because the format or the result MAY be different.
+
 The envelope MUST be signed with a symmetric key agreed between the requester and Mailserver.
 
 ### Receiving historic messages
 
-Historic messages MUST be sent to a peer with as packet with a P2P Message code (`0x7f`) followed by a single Whisper envelope OR an array of Whisper envelopes. A peer receiving historic message MUST handle both cases.
+Historic messages MUST be sent to a peer as a packet with a P2P Message code (`0x7f`) followed by an array of Whisper envelopes. It is incompatible with the original Whisper spec (EIP-627) because it allows only a single envelope, however, an array of envelopes is much more performant. In order to stay compatible with EIP-627, a peer receiving historic message MUST handle both cases.
 
-In order to accept a P2P Message packet, a node MUST trust a selected Mailserver.
+In order to receive historic messages from a mailserver, a node MUST trust the selected mailserver, that is allow to receive packets with the P2P Message code. By default, such packets are discarded.
 
 Received envelopes MUST be passed through the Whisper envelopes pipelines so that they are picked up by registered filters and passed to subscribers.
 
--- a/status-whisper-mailserver-spec.md
+++ b/status-whisper-mailserver-spec.md
@@ -47,7 +47,7 @@ In the Whisper envelope's payload section, there MUST be RLP-encoded information
 `Upper`: 4-byte wide unsigned integer (UNIX time in seconds; newest requested envelope's creation time)
 `Bloom`: 64-byte wide array of Whisper topics encoded in a bloom filter to filter envelopes
 `Limit`: 4-byte wide unsigned integer limiting the number of returned envelopes
-`Cursor`: 32-byte wide array of a cursor returned from the previous request (optional)
+`Cursor`: an array of a cursor returned from the previous request (optional)
 
 The `Cursor` field SHOULD be filled in if a number of envelopes between `Lower` and `Upper` is greater than `Limit` so that the requester can send another request using the obtained `Cursor` value. What exactly is in the `Cursor` is up to the implementation. The requester SHOULD NOT use a `Cursor` obtained from one mailserver in a request to another mailserver because the format or the result MAY be different.
 
@@ -60,6 +60,18 @@ Historic messages MUST be sent to a peer as a packet with a P2P Message code (`0
 In order to receive historic messages from a mailserver, a node MUST trust the selected mailserver, that is allow to receive packets with the P2P Message code. By default, such packets are discarded.
 
 Received envelopes MUST be passed through the Whisper envelopes pipelines so that they are picked up by registered filters and passed to subscribers.
+
+For a requester, to know that all messages have been sent by mailserver, it SHOULD handle P2P Request Complete code (`0x7d`). This code is followed by a a byte array with:
+
+```
+[ RequestID, LastEnvelopeHash, Cursor ]
+```
+
+`RequestID`: 32-byte wide array with a hash of the envelope containing request details
+`LastEnvelopeHash`: 32-byte wide array with a has of the last set envelope for the request
+`Cursor`: an array of a cursor returned from the previous request (optional)
+
+If `Cursor` is not empty, it means that not all messages were sent due to the set `Limit` in the request. One or more consecutive requests MAY be sent with `Cursor` field filled in in order to receive the rest of messages.
 
 ## Security considerations
 
--- a/status-whisper-mailserver-spec.md
+++ b/status-whisper-mailserver-spec.md
@@ -7,7 +7,8 @@
   - [Abstract](#abstract)
   - [Mailserver](#mailserver)
     - [Archiving messages](#archiving-messages)
-    - [Delivering messages](#delivering-messages)
+    - [Requesting messages](#requesting-messages)
+    - [Receiving historic messages](#receiving-historic-messages)
   - [Security considerations](#security-considerations)
     - [Confidentiality](#confidentiality)
     - [Altruistic and centralized operator risk](#altruistic-and-centralized-operator-risk)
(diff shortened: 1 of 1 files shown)


status-whisper-mailserver-spec.md
```diff
@@ -28,19 +29,35 @@ It is important to notice that Mailserver will only handle requests from its dir
 
 ### Archiving messages
 
-In order to store messages, one MUST implement the interface below and MUST register it within a Whisper service. The only known Whisper implementation that allows that is [geth](https://github.com/ethereum/go-ethereum).
+A node which wants to provide a mailserver capability MUST store envelopes incoming in messages packets identify by a code `0x01`. The envelopes can be stored in an arbitrary format, however, they MUST be serialized and deserialized to Whisper envelope format.
```
kdeme:
```suggestion
A node which wants to provide mailserver functionality MUST store envelopes from incoming messages packets (Whisper packet-code `0x01`). The envelopes can be stored in an arbitrary format, however, they MUST be serialized and deserialized to Whisper envelope format.
```
I prefer not to use the word capability as that is a term used in rlpx for subprotocols. This could create confusion

status-whisper-mailserver-spec.md
```diff
@@ -28,19 +29,35 @@ It is important to notice that Mailserver will only handle requests from its dir
...
 
-`MailServer` interface consist of:
-* `Archive(env *Envelope)`
-* `DeliverMail(whisperPeer *Peer, request *Envelope)`
+A mailserver SHOULD store envelopes for all topics to be useful by any peer, however, when for personal use case it MAY store only subset of topics.
```
kdeme:
```suggestion
A mailserver SHOULD store envelopes for all topics to be generally useful for any peer, however, for specific use cases it MAY store envelopes for a subset of topics.
```

status-whisper-mailserver-spec.md
```diff
@@ -28,19 +29,35 @@ It is important to notice that Mailserver will only handle requests from its dir
...
 
-### Delivering messages
+### Requesting messages
 
-Mailserver delivers archieved messages to a peer after receiving a Whisper packet with code `p2pRequestCode`. Messages are delivered asynchronously, i.e. a requester sends a Whisper packet with code `p2pRequestCode` and at some point later, it will start receiving Whisper packets with code `p2pMessageCode`.
+In order to request historic messages, a node MUST send a packet P2P Request (`0x7e`) to its peer providing mailserver capability. This packet requires one argument which MUST be a Whisper envelope.
```
kdeme:
```suggestion
In order to request historic messages, a node MUST send a packet P2P Request (`0x7e`) to its peer providing mailserver functionality. This packet requires one argument which MUST be a Whisper envelope.
```

status-whisper-mailserver-spec.md
```diff
@@ -28,19 +29,35 @@ It is important to notice that Mailserver will only handle requests from its dir
...
+[ Lower, Upper, Bloom, Limit, Cursor ]
+```
+
+`Lower`: 4-byte wide unsigned integer (UNIX time in seconds; oldest requested envelope's creation time)
+`Upper`: 4-byte wide unsigned integer (UNIX time in seconds; newest requested envelope's creation time)
+`Bloom`: 64-byte wide array of Whisper topics encoded in a bloom filter to filter envelopes
+`Limit`: 4-byte wide unsigned integer limiting the number of returned envelopes
+`Cursor`: 32-byte wide array of a cursor returned from the previous request (optional)
+
+The envelope MUST be signed with a symmetric key agreed between the requester and Mailserver.
```
kdeme, line 54: How is this symmetric key agreed up on?

status-whisper-mailserver-spec.md
```diff
@@ -28,19 +29,35 @@ It is important to notice that Mailserver will only handle requests from its dir
...
+
+### Receiving historic messages
+
+Historic messages MUST be sent to a peer with a packet with a P2P Message code (`0x7f`) followed by a single Whisper envelope OR an array of Whisper envelopes. A peer receiving historic message MUST handle both cases.
```
kdeme: This is a bit problematic. 
1. I believe it goes against the waku/0 or Whisper/6 spec as there the P2P message code is defined with a single envelope as parameter. 
2. In our nim implementation (DSL) it is not supported to do such thing (1 envelope or list of envelopes) .
So I would prefer to have waku/0 altered with either always a list of envelopes (can be 1). Or to add a message packet for multiple envelopes.

kdeme (commented):
Looks clear enough to me to implement. Did mostly some language suggestions (feel free to improve).

However, symmetric key item + the issue of the single versus list of envelopes are still open questions to me.

status-whisper-mailserver-spec.md
kdeme: Just realized that this does not need to be part of the spec and can be arranged out of band (layer higher, etc.)
However, should we then not leave the option open to use either a symmetric or asymmetric key? Such as with the regular posting of messages?

adambabik: Yes, I should make it more generic. It just must be encrypted somehow because each envelope requires that.

status-whisper-mailserver-spec.md
adambabik: > I believe it goes against the waku/0 or Whisper/6 spec as there the P2P message code is defined with a single envelope as parameter.

Correct but in the case when a mailserver returns a lot of messages, sending them one by one adds huge overhead. That's why we batched them.

> In our nim implementation (DSL) it is not supported to do such thing (1 envelope or list of envelopes) .
So I would prefer to have waku/0 altered with either always a list of envelopes (can be 1). Or to add a message packet for multiple envelopes.

I would opt for multiple envelopes then. If one wants to send just one, they can do that as one-item array. The problem is that it makes Waku incompatible with Whisper... Can we solve this in the bridge?

status-whisper-mailserver-spec.md
```diff
@@ -28,19 +29,35 @@ It is important to notice that Mailserver will only handle requests from its dir
 
 ### Archiving messages
 
-In order to store messages, one MUST implement the interface below and MUST register it within a Whisper service. The only known Whisper implementation that allows that is [geth](https://github.com/ethereum/go-ethereum).
+A node which wants to provide mailserver functionality MUST store envelopes from incoming messages packets (Whisper packet-code `0x01`). The envelopes can be stored in an arbitrary format, however, they MUST be serialized and deserialized to Whisper envelope format.
```
decanus:
```suggestion
A node which wants to provide mailserver functionality MUST store envelopes from incoming message packets (Whisper packet-code `0x01`). The envelopes can be stored in any format, however they MUST be serialized and deserialized to the Whisper envelope format.
```

status-whisper-mailserver-spec.md
```diff
@@ -28,19 +29,35 @@ It is important to notice that Mailserver will only handle requests from its dir
...
 
-`MailServer` interface consist of:
-* `Archive(env *Envelope)`
-* `DeliverMail(whisperPeer *Peer, request *Envelope)`
+A mailserver SHOULD store envelopes for all topics to be generally useful for any peer, however, for specific use cases it MAY store envelopes for a subset of topics.
```
decanus:
```suggestion
A mailserver SHOULD store envelopes for all topics to be generally useful for any peer, however for specific use cases it MAY store envelopes for a subset of topics.
```

status-whisper-mailserver-spec.md
```diff
@@ -28,19 +29,35 @@ It is important to notice that Mailserver will only handle requests from its dir
...
+`Upper`: 4-byte wide unsigned integer (UNIX time in seconds; newest requested envelope's creation time)
+`Bloom`: 64-byte wide array of Whisper topics encoded in a bloom filter to filter envelopes
+`Limit`: 4-byte wide unsigned integer limiting the number of returned envelopes
+`Cursor`: 32-byte wide array of a cursor returned from the previous request (optional)
+
+The envelope MUST be signed with a symmetric key agreed between the requester and Mailserver.
+
+### Receiving historic messages
+
+Historic messages MUST be sent to a peer with a packet with a P2P Message code (`0x7f`) followed by a single Whisper envelope OR an array of Whisper envelopes. A peer receiving historic message MUST handle both cases.
```
decanus:
```suggestion
Historic messages MUST be sent to a peer with as packet with a P2P Message code (`0x7f`) followed by a single Whisper envelope OR an array of Whisper envelopes. A peer receiving historic message MUST handle both cases.
```

decanus (changes requested):

decanus (approved):

status-whisper-mailserver-spec.md
```diff
@@ -28,19 +29,35 @@ It is important to notice that Mailserver will only handle requests from its dir
...
+`Limit`: 4-byte wide unsigned integer limiting the number of returned envelopes
+`Cursor`: 32-byte wide array of a cursor returned from the previous request (optional)
+
+The envelope MUST be signed with a symmetric key agreed between the requester and Mailserver.
+
+### Receiving historic messages
+
+Historic messages MUST be sent to a peer with as packet with a P2P Message code (`0x7f`) followed by a single Whisper envelope OR an array of Whisper envelopes. A peer receiving historic message MUST handle both cases.
+
+In order to accept a P2P Message packet, a node MUST trust a selected Mailserver.
```
oskarth: Can we be more specific about what we mean by "trust"? I suppose this means making certain envelope invariants / checks optional?

E.g. here is roughly the logic in nim-eth as far as I can tell https://github.com/status-im/nim-eth/blob/master/eth/p2p/rlpx_protocols/whisper_protocol.nim#L223-L225

status-whisper-mailserver-spec.md
```diff
@@ -28,19 +29,35 @@ It is important to notice that Mailserver will only handle requests from its dir
...
-Because all packets exchanged between a Mailserver and a peer are p2p packets, all filters created by a peer from which it expectes to receive archived messages MUST allow processing of direct peer-to-peer messages.
+```
+[ Lower, Upper, Bloom, Limit, Cursor ]
+```
+
+`Lower`: 4-byte wide unsigned integer (UNIX time in seconds; oldest requested envelope's creation time)
+`Upper`: 4-byte wide unsigned integer (UNIX time in seconds; newest requested envelope's creation time)
+`Bloom`: 64-byte wide array of Whisper topics encoded in a bloom filter to filter envelopes
+`Limit`: 4-byte wide unsigned integer limiting the number of returned envelopes
+`Cursor`: 32-byte wide array of a cursor returned from the previous request (optional)
```
oskarth: How is this returned exactly?

What does it correspond to? I suppose an address of some kind that can be used for that specific mailserver to skip previously fetched messages? Does it send messages 'before' this cursor or after? How does it impact Lower and Upper, i.e. does it cap it to be higher than Lower and lower than Higher according to some logic?

status-whisper-mailserver-spec.md
oskarth: Seems like a feature enhancement request to spec to me, I suggest we note that it has to be agreed out of band in implementer notes in waku spec

For Status app it is hardcoded, iirc.

status-whisper-mailserver-spec.md
```diff
@@ -28,19 +29,35 @@ It is important to notice that Mailserver will only handle requests from its dir
...
+
+The envelope MUST be signed with a symmetric key agreed between the requester and Mailserver.
+
+### Receiving historic messages
+
+Historic messages MUST be sent to a peer with as packet with a P2P Message code (`0x7f`) followed by a single Whisper envelope OR an array of Whisper envelopes. A peer receiving historic message MUST handle both cases.
```
oskarth: The single or array seems a bit suboptimal to me, ideally this should be just array and then it can even be empty, just like Messages packet.

Seems like an oversight in initial EIP627, perhaps this interface is something we can fix for later versions of waku.

oskarth (commented):
Much better than before, right level of abstraction now!

Some questions of clarifications on interface

status-whisper-mailserver-spec.md
kdeme: Yes, agree. That's why I suggested either changing this packet like you say -> array which can be empty. Or adding a new packet id, which does this, to the waku spec.

kdeme:
By the way, are any of the additional packet ids (still) used for mailserver functionality?

`p2pRequestComplete`, `p2pSyncRequest`, `p2pSyncResponse` sound related. (https://github.com/status-im/nim-eth/blob/master/eth/p2p/rlpx_protocols/whisper_protocol.nim#L228)

status-whisper-mailserver-spec.md
adambabik: Good point. Generally, a trusted peer in Whisper terminology means it's allowed to send p2p messages.

status-whisper-mailserver-spec.md
adambabik: It's up to the implementation. Also, it's not possible to query one mailserver and use the obtained cursor to query another one -- I will make it clear. It is so because each mailserver can obtain messages in different order or even may miss some. So it's Whisper property not this spec.

All other questions are up to the implementation due to what I wrote above.

status-whisper-mailserver-spec.md
adambabik: [I replied with the same suggestions](https://github.com/status-im/specs/pull/64#discussion_r349266498). Just note that it will be incompatible with EIP-627.

adambabik:
@kdeme `p2pSync*` are not relevant. They are used to sync between mailservers like when you start a new one but want to get history from another one. I would not worry about this now.

`p2pRequestComplete` might be actually worth to implement. Otherwise, it's unknown when all messages were sent from a mailserver. Unless we have a better idea for this.

status-whisper-mailserver-spec.md
adambabik: It is hardcoded, correct. I will keep it as it is then for now.

oskarth (approved):
LGTM

adambabik:
@kdeme I describe P2P Request Complete code in 4eb554a.

kdeme (approved):
OK for me now, however I think once we move to a Waku version of this spec we should readdress this:
`It is incompatible with the original Whisper spec (EIP-627) because it allows only a single envelope, however, an array of envelopes is much more performant. In order to stay compatible with EIP-627, a peer receiving historic message MUST handle both cases.`

No need to mention that much then anymore about compatibility with EIP-627.
It can be done in the section which explains the differences between Whisper and Waku somewhere?

adambabik:
@kdeme If we have a section about differences, we can reference it here. But I think we must mention what is incompatible for clarify in the relevant sections. I agree that explanation can be in a different place. 

> It can be done in the section which explains the differences between Whisper and Waku somewhere?

Waku is speced in a different repo. Here we just want to describe the current state of the protocol.

kdeme:
@adambabik  Yeah, for the current state of the protocol I'm perfectly fine with it. As we base ourselves on EIP-627 in those specs / implementations.

Just a reminder for later.

PR #68, opened 2019-11-26
Add EIP list

### LINKED ISSUE #6: Document specific EIPs supported
To avoid situations like this: https://discuss.status.im/t/sending-limit-per-contact/1179/6?u=oskarth

And make it clear which once we support and not.

Also see LES etc.

This is more of an INFORMATIONAL one for each client (status-go, Status App), but really useful to have at a glance.

3esmit:
Fixes #6

### DIFF
--- a/status-EIPs.md
+++ b/status-EIPs.md
@@ -25,11 +25,30 @@ Status should follow all standards as possible. Whenever a feature is needed, it
 
 ### Support table
 
-|           | BIP32 | BIP39 | BIP43 | BIP44 | EIP20 | EIP55 | EIP67 | EIP137 | EIP155 | EIP165 | EIP181 | EIP191 | EIP627 | EIP681 | EIP712 | EIP721 | EIP831 | EIP945 | EIP1102 | EIP1193 | EIP1577 | EIP1581 |
-| --------- | ----- | ----- | ----- | ----- | ----- | ----- | ----- | ------ | ------ | ------ | ------ | ------ | ------ | ------ | ------ | ------ | ------ | ------ | ------- | ------- | ------- | ------- |
-| Status v0 | N     | Y     | N     | N     | Y     | Y     | P     | P      | Y      | P      | P      | Y?     | Y      | Y      | P      | P      | N      | Y      | Y       | Y       | Y       | N       |
-| Status v1 | Y     | Y     | Y     | Y     | Y     | Y     | P     | P      | Y      | N      | N      | N      | Y      | N      | P      | P      | Y      | Y      | Y       | Y       | P       | Y       |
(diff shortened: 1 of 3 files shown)


3esmit:
There might be missing some EIPs. Feel free to ask for any changes regarding format.

oskarth:
This is awesome, thank you!

status-EIPs.md
```diff
@@ -0,0 +1,211 @@
+# Status EIPs standards
```
oskarth, line 1: Could we add a link to this in the README and from the main client spec as well?

oskarth (commented):
looks like a good start to me

status-EIPs.md
3esmit: Added to README. In main client spec I am not sure where this link fits.

3esmit:
I am not sure we want to call it EIP standards, but something more like "web3 standards", because it involves other then EIPs, such also BIPs and maybe in future other such as https://github.com/WalletConnect/WCIPs 

This document is intended to facilitate the interoperability of web3 in general, so describing all standards that Status follow. In case of EIPs, there is also some that were authored by Status, such as 1077 (gas abstraction), 1581 (non wallet keys) and 2400 (transaction receipt uri)

kdeme (commented):
Noticed when looking at the markdown formatted text that `Support:`, `Reference:`, etc. are not separated by newlines. I think you need to add a bunch of double spaces to do so. Or format it some other readable way.

decanus (approved):
this is great thanks!

status-EIPs.md
oskarth: Added commit with reference

status-EIPs.md
```diff
@@ -0,0 +1,211 @@
...
+  - [Security Considerations](#security-considerations)
+  - [Design Rationale](#design-rationale)
+  - [Footnotes](#footnotes)
+  - [Acknowledgements](#acknowledgements)
+
+## Introduction
+
+Status should follow all standards as possible. Whenever a feature is needed, it should be first checked if there is a standard for that, if not, Status should propose a standard.
+
+### Support table
```
oskarth, line 26: A bit hard to horizontally scroll, I wonder if it makes sense to flip these axes around?

oskarth (approved):
Fixed the client spec mention, I'd be fine to merge this

status-EIPs.md
3esmit: Good idea. Just transposed the table. I used this https://www.tablesgenerator.com/markdown_tables

kdeme (approved):

PR #74, opened 2019-12-08
Add message confirmations spec

### LINKED ISSUE #70: Document how message confirmations work
As a user,
I would like to know whether my message is properly sent through the wire and received by peers
so that I can be sure it has a chance to be propagated and received by others.

## Acceptance criteria

* Describe the current mechanism of message confirmations.
* Document all packet codes used to achieve this.
* Describe the role of peers in message confirmations.

adambabik:
Closes #70

### DIFF
--- a/status-whisper-usage-spec.md
+++ b/status-whisper-usage-spec.md
@@ -20,6 +20,7 @@
     - [One-to-one topic](#one-to-one-topic)
     - [Group chat topic](#group-chat-topic)
   - [Message encryption](#message-encryption)
+  - [Message confirmations](#message-confirmations)
   - [Whisper V6 extensions](#whisper-v6-extensions)
     - [Request historic messages](#request-historic-messages)
       - [shhext_requestMessages](#shhextrequestmessages)
@@ -244,6 +245,29 @@ Even though, the protocol specifies an encryption layer that encrypts messages b
 Public and group messages are encrypted using symmetric encryption and the key is created from a channel name string. The implementation is available in [`shh_generateSymKeyFromPassword`](https://github.com/ethereum/go-ethereum/wiki/Whisper-v6-RPC-API#shh_generatesymkeyfrompassword) JSON-RPC method of go-ethereum Whisper implementation.
 
 One-to-one messages are encrypted using asymmetric encryption.
+
+## Message confirmations
+
+Sending a message is a complex process where many things can go wrong. Message confirmations tells a node that a message originating from it has been received by its peers.
+
+A node MAY send a message confirmation for any batch of messages received with a packet Messages Code (`0x01`).
+
+A message confirmation is sent using Batch Acknowledge packet (`0x0b`) or Message Response packet (`0x0c`).
+
+The Batch Acknowledge packet is followed by a keccak256 hash of the envelopes batch data (raw bytes).
+
+The Message Response packet is more complex and is followed by a Versioned Message Response:
+```
+[ Version, Response]
+```
+
+`Version`: a version of the Message Response, equal to `1`,
+`Response`: `[ Hash, Errors ]` where `Hash` is a keccak256 hash of the envelopes batch data (raw bytes) for which the confirmation is sent and `Errors` is a list of envelope errors when processing the batch. A single error contains `[ Hash, Code, Description ]` where `Hash` is a hash of the processed envelope, `Code` is an error code and `Description` is a descriptive error message.
+
+The supported codes:
+`1`: means time sync error which happens when an envelope is too old or created in the future (the root cause is no time sync between nodes).
+
+The drawback of sending message confirmations is that it increases the noise in the network because for each sent message, a corresponding confirmation is broadcasted by one or more peers.
 
 ## Whisper V6 extensions
 
--- a/status-whisper-usage-spec.md
+++ b/status-whisper-usage-spec.md
@@ -248,7 +248,7 @@ One-to-one messages are encrypted using asymmetric encryption.
 
 ## Message confirmations
 
-Sending a message is a complex process where many things can go wrong. Message confirmations tells a node that a message originating from it has been received by its peers.
+Sending a message is a complex process where many things can go wrong. Message confirmations tell a node that a message originating from it has been received by its peers.
 
 A node MAY send a message confirmation for any batch of messages received with a packet Messages Code (`0x01`).
 


status-whisper-usage-spec.md
```diff
@@ -245,6 +246,29 @@ Public and group messages are encrypted using symmetric encryption and the key i
 
 One-to-one messages are encrypted using asymmetric encryption.
 
+## Message confirmations
+
+Sending a message is a complex process where many things can go wrong. Message confirmations tells a node that a message originating from it has been received by its peers.
```
decanus:
```suggestion
Sending a message is a complex process where many things can go wrong. Message confirmations tell a node that a message originating from it has been received by its peers.
```

decanus (changes requested):

status-whisper-usage-spec.md
oskarth: what does "received by its peers" mean? Which peers? Wouldn't this result in a lot of extra traffic?

status-whisper-usage-spec.md
```diff
@@ -245,6 +246,29 @@ Public and group messages are encrypted using symmetric encryption and the key i
 
 One-to-one messages are encrypted using asymmetric encryption.
 
+## Message confirmations
```
oskarth, line 249: This API in general is quite confusion to me. Is there any history of how this came to be? What's the rationale? Why does it do ad hoc message sync when we have MVDS for that on top? Doesn't it lead to a lot of extra traffic? Do we definitely want to keep it? Do we dependent on it?

oskarth (commented):
Initial comments

status-whisper-usage-spec.md
```diff
@@ -245,6 +246,29 @@ Public and group messages are encrypted using symmetric encryption and the key i
...
+
+Sending a message is a complex process where many things can go wrong. Message confirmations tells a node that a message originating from it has been received by its peers.
+
+A node MAY send a message confirmation for any batch of messages received with a packet Messages Code (`0x01`).
+
+A message confirmation is sent using Batch Acknowledge packet (`0x0b`) or Message Response packet (`0x0c`).
```
oskarth, line 255: Whats their int packet code?

status-whisper-usage-spec.md
```diff
@@ -245,6 +246,29 @@ Public and group messages are encrypted using symmetric encryption and the key i
...
+
+The Batch Acknowledge packet is followed by a keccak256 hash of the envelopes batch data (raw bytes).
+
+The Message Response packet is more complex and is followed by a Versioned Message Response:
+```
+[ Version, Response]
+```
```
oskarth, line 262: I suppose this is ok for here, but for Waku lets clean this up with a more concise ABNF grammar imo cc @decanus (i.e. surrounding text)

status-whisper-usage-spec.md
```diff
@@ -245,6 +246,29 @@ Public and group messages are encrypted using symmetric encryption and the key i
...
+
+`Version`: a version of the Message Response, equal to `1`,
+`Response`: `[ Hash, Errors ]` where `Hash` is a keccak256 hash of the envelopes batch data (raw bytes) for which the confirmation is sent and `Errors` is a list of envelope errors when processing the batch. A single error contains `[ Hash, Code, Description ]` where `Hash` is a hash of the processed envelope, `Code` is an error code and `Description` is a descriptive error message.
+
+The supported codes:
+`1`: means time sync error which happens when an envelope is too old or created in the future (the root cause is no time sync between nodes).
```
oskarth, line 268: Not sure I follow

status-whisper-usage-spec.md
```diff
@@ -245,6 +246,29 @@ Public and group messages are encrypted using symmetric encryption and the key i
...
+
+The drawback of sending message confirmations is that it increases the noise in the network because for each sent message, a corresponding confirmation is broadcasted by one or more peers.
```
oskarth, line 270: It seems like a huge amplification factor. If you send one message, and you are connected to N peers you would get N extra envelopes in the network?

What's our current policy for using this? Client-Cluster nodes, just inside cluster, etc?

status-whisper-usage-spec.md
adambabik: Node's peers. I think it's pretty clear in this context -- each node has its own peers (other nodes it is connected to).

> Wouldn't this result in a lot of extra traffic?

Yes, it might increase the traffic quite a lot. Currently, only mailservers sent confirmations. If one sends N batches, and there are M nodes sending confirmations, we have N*M additional packets floating around. They are not broadcasted though, only one hop.

status-whisper-usage-spec.md
adambabik: A lot of questions :D 

> Is there any history of how this came to be? What's the rationale?

I don't recall anything, I haven't participated in the implementation, I don't know why we have to packet codes.

> Why does it do ad hoc message sync when we have MVDS for that on top?

This was implemented much earlier than MVDS. How does MVDS solve this problem? The crucial thing is to know whether another peer actually received the message. In our case, mailservers send confirmations so it's highly likely that the message was propagated properly among all nodes.

> Doesn't it lead to a lot of extra traffic?

It adds some. I explained in a different comment. It's important to remember this packet is not broadcasted, there is only one hope for the message confirmation packets.

> Do we definitely want to keep it? Do we dependent on it?

It depends on the app requirements. I think currently it is required to display the status of a message.

status-whisper-usage-spec.md
adambabik: `11` and `12`. What's not clear with hex?

status-whisper-usage-spec.md
adambabik: `Code` in the paragraph above can have the following values. The only available option is `1`. 
Should it say `the root cause is lack of proper time synchronization between nodes`?

status-whisper-usage-spec.md
oskarth: Ack. Do we have any statistics for how big this overhead is in terms of BW?

status-whisper-usage-spec.md
kdeme: I too find this quite a strange addition to the API. 
- I don't understand how this actually confirms anything. If used only for 1 hop, a confirmation does not give you any guarantee of arrival to the real recipients. 
- It would introduce a whole lot of traffic when applied on all nodes (for each hop again). (I read that it is only active for Mail servers ? )

So I guess it is now just used for mail servers so you are sure that your message gets in a mailserver database there, but only if you are directly connected with those, which in the current setup, I guess you are. Thisis the kind of functionality, that is really based too much on the network setup imo.

Also, some thought regarding privacy and security implications would be wise here.

Anyway, I understand that this is the situation as is, but for Waku/0 or Waku/1 we should definitely revise whether this is still needed. And not already handled by the layer up (as I believe it should be when a transport as Whisper is used).

status-whisper-usage-spec.md
adambabik: What is BW? 

No, we don't have any stats. It's not a big deal from a client perspective. For each batch of envelopes, the client receives a single, quite small packet. The only thing which is not necessary is the fact that a node does not know whether a peer wants the confirmation at all. This can be improved.

status-whisper-usage-spec.md
adambabik: > I don't understand how this actually confirms anything. If used only for 1 hop, a confirmation does not give you any guarantee of arrival to the real recipients.

No, it only gives you a confirmation that a message has been received by at least one node. In our case, this node is mailserver.

> It would introduce a whole lot of traffic when applied on all nodes (for each hop again).

That's why it's not broadcast-based.

> Thisis the kind of functionality, that is really based too much on the network setup imo.

Generally yes. It assumes that you are directly connected to a mailserver and actually if you want to use mailserver, you have to be directly connected to it.

> Anyway, I understand that this is the situation as is, but for Waku/0 or Waku/1 we should definitely revise whether this is still needed. And not already handled by the layer up (as I believe it should be when a transport as Whisper is used).

Correct. Like I mentioned in some other comment, this solution was introduced rather quickly and without consulting enough people. It's definitely not something we should be attached to. 

This is more general problem of storing offline messages. The granularity of confirmations might be also increased to improve the UX.

oskarth:
Ping @adambabik

status-whisper-usage-spec.md
adambabik: I answered in the other comments. Currently confirmations are sent only by mailservers directly to a peer from which a batch of messages was received.

adambabik:
@oskarth I answered the comments, waiting for approvals. I don't have an intention to add more changes for now.

oskarth:
Ok, if you need people to look it over again and people aren't, please ping. It's your PR :)

status-whisper-usage-spec.md
oskarth: BW is bandwidth

status-whisper-usage-spec.md
oskarth: As long as we are consistent, it can lead to confusion if we aren't careful about it IMO. See e.g. https://specs.vac.dev/waku.html and packet codes

status-whisper-usage-spec.md
oskarth: Please ack for Vac @decanus

status-whisper-usage-spec.md
oskarth: Sure

oskarth (approved):
Ok to merge, see other comments. Can iterate and also fix some thing Waku spec

oskarth:
@decanus @kdeme re-requested reviews on Adam's behalf

kdeme (approved):
OK for me.
My comments were rather to take into account for possible future changes with regards to the Waku spec.

status-whisper-usage-spec.md
dshulyak: i can try to explain, status chat notifies user when message was sent to the network. and user is able to retry only if message failed to be ack'ed by the network.

the rationale was that you never know if the message was actually sent to the network or not, socket api returns immediately when message was written to a socket buffer. so there was a ton of false positives and this solution was added.

mvds gives you confirmation that the message was actually delivered to the end recipient and it is a much better guarantee then this solution, but not quite the same. you can think of it like a first confirmation from telegram server, which is obviously a different model.

as for bw confirmation was made for whole batch of whisper envelopes, so this is like 32 bytes overhead per single batch. not broadcasted like it was noted by Adam. just 32 bytes directly written to a sender of the batch. pretty sure that amount is negligible in comparison with other whisper traffic

dshulyak: btw i am not advocating for this solution to stay or smth like that and perfectly understand that it provides some guarantees only because of how status cluster work. just wanted to provide some intuition on how it came to be.

PR #79, opened 2020-01-16
eip-1459

decanus:
This PR contains everything related to EIP-1459 which will be used to bootstrap for waku.


status-EIPs.md
```diff
@@ -220,6 +221,14 @@ Used for: Security (dont reuse wallet key) and user experience (dont request key
 Related: https://github.com/status-im/status-react/issues/9088 https://github.com/status-im/status-react/pull/9096  
 Sourcecode: https://github.com/status-im/status-react/blob/develop/src/status_im/constants.cljs#L242  
 
+### EIP1459 - Node Discovery via DNS
+
+Support: -
+Reference: https://eips.ethereum.org/EIPS/eip-1459
+Description: Allows the storing and retreving of nodes through merkle trees stored in TXT records of a domain.
+Used for: Finding Waku nodes.
```
corpetty: should this be `Waku` nodes?

oskarth: EIP-1459 is used to find bootstrap nodes, do we definitely want to use it to hardcode Waku nodes too? Does it make sense to separate these?

also cc @adambabik

oskarth:
This PR should also update the current bootstrap section:

https://github.com/status-im/specs/blob/master/status-client-spec.md#bootstrapping

And possibly discovery section:

https://github.com/status-im/specs/blob/master/status-client-spec.md#discovery

This is where it'd be very useful to have a v1 tag, because then the spec diff is clear

oskarth (commented):
See initial comments

status-EIPs.md
adambabik: If we want to have a fully-fledged peers discovery mechanism then we would put there only bootnodes which would provided waku nodes.

The options we have:
1. Hardcoded static list of Waku nodes -- changing it requires a new app release,
2. A list of nodes taken from https://fleets.status.im/ -- periodically update the cached list of nodes,
3. Waku nodes in DNS -- like (2) but DNS based instead of HTTPS,
4. Bootnodes in DNS and a peers discovery mechanism.

In reasonable amount of time, only 1-3 are possible. (2) seems like the easiest.

decanus:
@oskarth see changed to the client spec, short and sweet for now.

status-EIPs.md
decanus: I prefer `3`, so that this is the bootstrap mechanism fully. @oskarth what are your opinions? I wrote it similar to option `2` in the spec right now however.

decanus: @cammellos opinion appreciated here too.

cammellos: 3 is also my preferred option, the only thing is that I would have 3 ideally as an additional (and preferred) mechanism to 1.

If 3 fails for whatever reason (DNS is filtered, DNS entry is corrupted/not available etc), we fallback on a static IP address.

That also helps with the fact that if a user wants to set up they own bootnodes, they might not add a DNS entry and use plain IP addresses for that, so having both mechanism is desirable.

Essentially: 
definitely 3 , ideally 3 with fallback/in addition to 1

oskarth: Gracefully I'd do it as follows:

a) 3 now with 1 as fall back / alternative 
b) Ensure there's a clean path to 4, since that's a more permissionless and proper setup with Discovery v5 (or similar). In that world, DNS would only host bootstrap nodes, and Waku nodes would be discovered (and likely persisted once found)

oskarth: I'd clarify this to say:
- used for finding bootstrap nodes

As well as currently:
- to find Waku nodes

Then once we get to integrate more proper support for Discovery v5 (or similar), we can adjust spec

status-client-spec.md
```diff
@@ -107,19 +107,8 @@ nodes allow you to discover other nodes of the network.
...
-- Europe:
-  - `enode://436cc6f674928fdc9a9f7990f2944002b685d1c37f025c1be425185b5b1f0900feaf1ccc2a6130268f9901be4a7d252f37302c8335a2c1a62736e9232691cc3a@174.138.105.243:443`
-  - `enode://5395aab7833f1ecb671b59bf0521cf20224fe8162fc3d2675de4ee4d5636a75ec32d13268fc184df8d1ddfa803943906882da62a4df42d4fccf6d17808156a87@206.189.243.57:443`
-- North America:
-  - `enode://7427dfe38bd4cf7c58bb96417806fab25782ec3e6046a8053370022cbaa281536e8d64ecd1b02e1f8f72768e295d06258ba43d88304db068e6f2417ae8bcb9a6@104.154.88.123:443`
-  - `enode://ebefab39b69bbbe64d8cd86be765b3be356d8c4b24660f65d493143a0c44f38c85a257300178f7845592a1b0332811542e9a58281c835babdd7535babb64efc1@35.202.99.224:443`
-
-These bootstrap nodes do not change, however, we can't guarantee that it will stay this way forever
-and at some point we might be forced to change them.
+In order to retrieve the addresses of bootstrap nodes, a client MUST traverse
```
oskarth: inconsistent with above, here it says bootstrap nodes vs also waku nodes above

status-client-spec.md
```diff
@@ -107,19 +107,8 @@ nodes allow you to discover other nodes of the network.
 Currently the main bootstrap nodes are provided by Status Gmbh, but anyone can
 run these provided they are connected to the rest of the Whisper network.
 
-Status maintains a list of boootstrap nodes in the following locations:
-- Asia:
```
oskarth: In line with above discussion, these should be fallback. I'd put them in an appendix and specify MAY or SHOULD specify fallback mechanism for bootstrap nodes in case DNS has issues, etc. It doesn't belong in main spec imo.

status-client-spec.md
```diff
@@ -107,19 +107,8 @@ nodes allow you to discover other nodes of the network.
...
-- North America:
-  - `enode://7427dfe38bd4cf7c58bb96417806fab25782ec3e6046a8053370022cbaa281536e8d64ecd1b02e1f8f72768e295d06258ba43d88304db068e6f2417ae8bcb9a6@104.154.88.123:443`
-  - `enode://ebefab39b69bbbe64d8cd86be765b3be356d8c4b24660f65d493143a0c44f38c85a257300178f7845592a1b0332811542e9a58281c835babdd7535babb64efc1@35.202.99.224:443`
-
-These bootstrap nodes do not change, however, we can't guarantee that it will stay this way forever
-and at some point we might be forced to change them.
+In order to retrieve the addresses of bootstrap nodes, a client MUST traverse
+the merkle tree found at [`fleet.status.im`](https://fleet.status.im), as described in [EIP-1459](https://eips.ethereum.org/EIPS/eip-1459#client-protocol).
 
 #### Discovery
```
oskarth: Not in diff, but this section

> To implement a Status client you need to discover peers to connect to. We use a light discovery mechanism based on a combination of Discovery v5 and Rendezvous Protocol, (with some modifications). Additionally, some static nodes MAY also be used.

Should be rewritten. We have to be a bit careful about qualifiers here as the spec is moving, so one idea is to approach it as follows:

1) Status v1, current spec reflecting app v1
2) v1.1 with DNS+Waku as option for bootstrap/waku nodes
3) v1.2 (or later) re-introduce proper Discovery and drop support for above hacky disc v5 and rendezvous

This means we have to be mindful of how we use MAY, SHOULD, MUST in terms of when we introduce capabilities and when we deprecate them. For example, the above MUST for DNS isn't accurate, it is a SHOULD, as the current Status v1 spec is otherwise not accurate. Etc.

oskarth (changes requested):

oskarth:
Even if we happen to use the same implementation mechanism for bootstrap and general discovery, lets make sure we keep these sections separate, as they are separate components. Us using the same mechanism is a convenience thing.

status-EIPs.md
adambabik: Ok, to fully understand: the preferred option is to put Waku nodes in DNS. In the future, we will seamlessly change them to bootnodes and status-go implementation accordingly. We also want to have a mechanism to load Waku nodes from a static file which will be periodically updated in status-go from https://fleets.status.im/. There should be a way to pass a file as a text or as an HTTPS URL, both provided as bindings to status-react.

adambabik: But we only support Waku nodes in DNS for now. If one puts a bootstrap node in there, they will be ignored. In the future, we can change it and even have a mix of nodes. This will be handled fine as long as a node is a valid enode address.

decanus:
So @fjl and I had an interesting discussion in Berlin today, about all the tooling they use for example to build the merkle tree etc. What is interesting is they crawl the DHT to find the nodes which they later add to the bootstrapping, decided to just note this here and make @fjl aware of the fact that we are using this issue.

status-EIPs.md
decanus: @adambabik I don't see the use of adding waku nodes into the static file anymore if we use DNS discovery.

oskarth (approved):
lgtm, seems like consensus is to do the bootstrap node list thing later

status-EIPs.md
```diff
@@ -220,6 +221,14 @@ Used for: Security (dont reuse wallet key) and user experience (dont request key
 Related: https://github.com/status-im/status-react/issues/9088 https://github.com/status-im/status-react/pull/9096  
 Sourcecode: https://github.com/status-im/status-react/blob/develop/src/status_im/constants.cljs#L242  
 
+### EIP1459 - Node Discovery via DNS
+
+Support: -
+Reference: https://eips.ethereum.org/EIPS/eip-1459
+Description: Allows the storing and retreving of nodes through merkle trees stored in TXT records of a domain.
```
kdeme: Typo
```suggestion
Description: Allows the storing and retrieving of nodes through merkle trees stored in TXT records of a domain.
```

kdeme (approved):

status-client-spec.md
```diff
@@ -169,6 +169,9 @@ new ones.
 The current list of static peers is published on https://fleets.status.im/. `eth.beta` is the current
 group of peers the official Status client uses. The others are test networks.
 
+Finally, Waku node addresses can be retrieved by traversing
```
adambabik: The consequence of this is adding Waku nodes as DNS TXT records to https://fleets.status.im. cc @jakubgs

status-client-spec.md
```diff
@@ -169,6 +169,9 @@ new ones.
 The current list of static peers is published on https://fleets.status.im/. `eth.beta` is the current
 group of peers the official Status client uses. The others are test networks.
 
+Finally, Waku node addresses can be retrieved by traversing 
+the merkle tree found at [`fleet.status.im`](https://fleet.status.im), as described in [EIP-1459](https://eips.ethereum.org/EIPS/eip-1459#client-protocol).
```
adambabik:
```suggestion
the merkle tree found at [`fleets.status.im`](https://fleets.status.im), as described in [EIP-1459](https://eips.ethereum.org/EIPS/eip-1459#client-protocol).
```

status-client-spec.md
adambabik: What we currently did in `eth.test` is that we run Waku and Whisper subprotocols on the same nodes. So really no additional discovery is needed to start integrating with the app.

jakubgs: If we want to use DNS records we can include those in https://fleets.status.im/.

decanus: @adambabik I was using it as a placeholder we can do it with whatever subdomain.

oskarth:
Ping @decanus @adambabik what's needed to get this merged?

decanus:
@oskarth I don't believe its actually implemented anywhere. Once it is I am happy to merge.

adambabik (approved):
LGTM

adambabik:
It's not implemented anywhere yet, for now I am trying to upgrade to geth 1.9.11 (https://github.com/status-im/status-go/pull/1899) as a prerequisite.

status-client-spec.md
```diff
@@ -169,6 +169,9 @@ new ones.
 The current list of static peers is published on https://fleets.status.im/. `eth.beta` is the current
```
jakubgs: It's actually `eth.prod`.

status-EIPs.md
```diff
@@ -49,6 +49,7 @@ Status should follow all standards as possible. Whenever a feature is needed, it
 | EIP1193  | Y         | Y         | Y        |
 | EIP1577  | Y         | P         | N        |
 | EIP1581  | N         | Y         | N        |
+| EIP1459  | N         |           | N        |
```
oskarth: We need to clarify situation wrt v1 stable and this being added in v1.2

oskarth (changes requested):

oskarth:
@decanus IIRC consensus was to merge this as raw draft, could you adjust accordingly please?

decanus:
@oskarth so the problem we still had which I remember is how to merge, because the EIP sped is marked as final, but now we are adding a draft thing to a final spec.

oskarth:
The solution proposed IIRC was to add a column indicating "raw" state. This document is more of a compilation than a spec on its own.

oskarth (approved):
LGTM

@andremedeiros @yenda could you sanity check that this reflects current state of app?

decanus:
ping @andremedeiros @yenda

yenda (approved):

PR #80, opened 2020-01-20
replace transit with protobuf

cammellos:
Updates spec to protobuf.

### DIFF
--- a/status-client-spec.md
+++ b/status-client-spec.md
@@ -69,7 +69,7 @@ Implementing a Status clients means implementing the following layers. Additiona
 
 ### Protobuf
 
-We use `protobuf` in different layers, the version used is `proto3` unless stated otherwise.
+We use [`protobuf`](https://developers.google.com/protocol-buffers/) in different layers, the version used is `proto3` unless stated otherwise.
 
 ## Components
 
--- a/status-client-spec.md
+++ b/status-client-spec.md
@@ -67,6 +67,10 @@ Implementing a Status clients means implementing the following layers. Additiona
 | Transport privacy | Routing, Metadata protection    | Whisper                      |
 | P2P Overlay       | Overlay routing, NAT traversal  | devp2p                       |
 
+### Protobuf
+
+We use `protobuf` in different layers, the version used is `proto3` unless stated otherwise.
+
 ## Components
 
 ### P2P Overlay
--- a/status-group-chats-spec.md
+++ b/status-group-chats-spec.md
@@ -28,177 +28,123 @@ This documents describes the group chat protocol used by the status application.
 
 ## Membership updates
 
-Membership updates messages are used to propagate group chat membership changes. The transit format is described in the [Status Payload Specs](status-payload-specs.md). Here we will be describing each specific field.
+Membership updates messages are used to propagate group chat membership changes. The protobuf format is described in the [Status Payload Specs](status-payload-specs.md). Here we will be describing each specific field.
 
-The format is:
+The protobuf messages are:
 
-```
-{
-  "events": [struct {"type": string, "member": string, "members": [string], "clock-value": uint, "name": string],
-  "signature": string,
-  "chat-id": string
+```protobuf
+// MembershipUpdateMessage is a message used to propagate information
+// about group membership changes.
+message MembershipUpdateMessage {
+  // The chat id of the private group chat
+  string chat_id = 1;
+  // A list of events for this group chat, first 65 bytes are the signature, then is a 
+  // protobuf encoded MembershipUpdateEvent
+  repeated bytes events = 2;
+  // An optional chat message
+  ChatMessage message = 3;
+}
+
+message MembershipUpdateEvent {
+  // Lamport timestamp of the event as described in [Status Payload Specs](status-payload-specs.md#clock-vs-timestamp-and-message-ordering)
+  uint64 clock = 1;
+  // List of public keys of the targets of the action
+  repeated string members = 2;
+  // Name of the chat for the CHAT_CREATED/NAME_CHANGED event types
+  string name = 3;
+  // The type of the event
+  EventType type = 4;
+
+  enum EventType {
+    UNKNOWN = 0;
+    CHAT_CREATED = 1; // See [CHAT_CREATED](#chat-created)
+    NAME_CHANGED = 2; // See [NAME_CHANGED](#name-changed)
+    MEMBERS_ADDED = 3; // See [MEMBERS_ADDED](#members-added)
+    MEMBER_JOINED = 4; // See [MEMBER_JOINED](#member-joined)
+    MEMBER_REMOVED = 5; // See [MEMBER_REMOVED](#member-removed)
+    ADMINS_ADDED = 6; // See [ADMINS_ADDED](#admins-added)
+    ADMIN_REMOVED = 7; // See [ADMIN_REMOVED](#admin-removed)
+  }
 }
 ```
 
+### Payload
+
+`MembershipUpdateMessage`:
+
+| Field | Name | Type | Description |
+| ----- | ---- | ---- | ---- |
+| 1 | chat-id | `string` | The chat id of the chat where the change is to take place |
+| 2 | events | See details | A list of events that describe the membership changes, in their encoded protobuf form |
+| 3 | message | `ChatMessage` | An optional message, described in [Message](#message) |
+
+`MembershipUpdateEvent`:
+
+| Field | Name | Type | Description |
+| ----- | ---- | ---- | ---- |
+| 1 | clock | `uint64` | The clock value of the event |
+| 2 | members | `[]string` | An optional list of hex encoded (prefixed with `0x`) public keys, the targets of the action |
+| 3 | name | `name` | An optional name, for those events that make use of it |
+| 4 | type | `EventType` | The type of event sent, described below |
+
+
 ### Chat ID
 
-Each membership update MUST be sent with a corresponding `chat-id`. 
-The format of this chat id MUST be a string, [UUID](https://tools.ietf.org/html/rfc4122 ), concatenated with the hex-encoded public key of the creator of the chat. This chat-id MUST be validated by all clients, and MUST be discarded if it does not follow these rules.
+Each membership update MUST be sent with a corresponding `chatId`. 
+The format of this chat ID MUST be a string of [UUID](https://tools.ietf.org/html/rfc4122 ), concatenated with the hex-encoded public key of the creator of the chat, joined by `-`. This chatId MUST be validated by all clients, and MUST be discarded if it does not follow these rules.
 
 ### Signature
 
-The signature for each event is calculated by creating a JSON array of all the `events` sorted by `clock-value` in ascending order, where each event is transformed in an array of tuples `field-name`, `value`, sorted by `field-name` in ascending alphabetical order. The last element of the array MUST be the `chat-id`. 
-Empty fields MUST be removed.
-
-For example the event:
-
-```
-   {
-     "chat-id": "chat-id",
-       "events":  [
-       {"b": "b-value"
-        "clock-value": 1,
-        "a": "a-value"
-        },
-       {
-         "e": "e-value",
-         "clock-value": 0,
-         "a": "a-value"
-       }
-       ]
-   }
-
-```
-
-Results in the structure:
-
-```
-   [
-     [
-       [
-         ["a" "a-value"],
-         ["clock-value", 0],
-         ["e" "e-value"]
-       ],
-       [
-         ["a", "a-value"],
-         ["b", "b-value"],
-         ["clock-value", 1]
-       ]
-     ],
-     "chat-id"
-   ]
-```
-
-This structure is then stringified collapsing all whitespaces and the `Keccak256` of the string is then signed using its private key by the author and added to the payload.
+The signature for each event is calculated by encoding each `MembershipUpdateEvent` in its protobuf representation and prepending the bytes of the chatID, lastly the `Keccak256` of the bytes is signed using the private key by the author and added to the `events` field of MembershipUpdateMessage.
       
-
 ### Group membership event
 
 Any group membership event received MUST be verified by calculating the signature as per the method described above. 
 The author MUST be extracted from it, if the verification fails the event MUST be discarded.
 
-#### chat-created
-
-```
-{
-  "type": "chat-created",
-  "name": string
-  "clock-value": uint
-}
-```
-
+#### CHAT_CREATED
 
 Chat created event is the first event that needs to be sent. Any event with a clock value lower then this MUST be discarded.
-Upon receiving this event a client MUST validate the `chat-id` provided with the updates and create a chat with identified by `chat-id` and named `name`.
+Upon receiving this event a client MUST validate the `chatId` provided with the updates and create a chat with identified by `chatId` and named `name`.
 
-#### name-changed
-
-```
-{
-  "type": "name-changed"
-  "name": string
-  "clock-value": uint
-}
-```
+#### NAME_CHANGED
 
 A name changed event is used by admins to change the name of the group chat.
-Upon receiving this event a client MUST validate the `chat-id` provided with the updates and MUST ensure the author of the event is an admin of the chat, otherwise the event MUST be ignored.
+Upon receiving this event a client MUST validate the `chatId` provided with the updates and MUST ensure the author of the event is an admin of the chat, otherwise the event MUST be ignored.
 If the event is valid the chat name SHOULD be changed to `name`.
 
-
-#### members-added
-
-```
-{
-  "type": "members-added"
-  "members": [string]
-  "clock-value": uint
-}
-```
+#### MEMBERS_ADDED
 
 A members added event is used by admins to add members to the chat.
-Upon receiving this event a client MUST validate the `chat-id` provided with the updates and MUST ensure the author of the event is an admin of the chat, otherwise the event MUST be ignored.
+Upon receiving this event a client MUST validate the `chatId` provided with the updates and MUST ensure the author of the event is an admin of the chat, otherwise the event MUST be ignored.
 If the event is valid a client MUST update the list of members of the chat who have not joined, adding the `members` received.
 `members` is an array of hex encoded public keys.
 
-#### member-joined
-
-```
-{
-  "type": "member-joined"
-  "member": string
-  "clock-value": uint
-}
-```
+#### MEMBER_JOINED
 
 A members joined event is used by a member of the chat to signal that they want to start receiving messages from this chat.
-Upon receiving this event a client MUST validate the `chat-id` provided with the updates and MUST ensure the author of the event is the same as the one specified by the `member` field.
-If the event is valid a client MUST update the list of members of the chat who joined, adding `member`. Any `message` sent to the group chat should now include the newly joined member.
+Upon receiving this event a client MUST validate the `chatId` provided with the updates.
+If the event is valid a client MUST update the list of members of the chat who joined, adding the signer. Any `message` sent to the group chat should now include the newly joined member.
 
-#### admins-added
-
-```
-{
-  "type": "admins-added"
-  "members": [string]
-  "clock-value": uint
-}
-```
+#### ADMINS_ADDED
 
 An admins added event is used by admins to add make other admins in the chat.
-Upon receiving this event a client MUST validate the `chat-id` provided with the updates, MUST ensure the author of the event is an admin of the chat and MUST ensure all `members` are already `members` of the chat, otherwise the event MUST be ignored.
+Upon receiving this event a client MUST validate the `chatId` provided with the updates, MUST ensure the author of the event is an admin of the chat and MUST ensure all `members` are already `members` of the chat, otherwise the event MUST be ignored.
 If the event is valid a client MUST update the list of admins of the chat, adding the `members` received.
 `members` is an array of hex encoded public keys.
 
-#### member-removed
-
-```
-{
-  "type": "member-removed"
-  "member": string
-  "clock-value": uint
-}
-```
+#### MEMBER_REMOVED
 
 A member-removed event is used to leave or kick members of the chat.
-Upon receiving this event a client MUST validate the `chat-id` provided with the updates, MUST ensure that:
+Upon receiving this event a client MUST validate the `chatId` provided with the updates, MUST ensure that:
 - If the author of the event is an admin, target can only be themselves or a non-admin member.
 - If the author of the event is not an admin, the target of the event can only be themselves.
 -
 If the event is valid a client MUST remove the member from the list of `members`/`admins` of the chat, and no further message should be sent to them.
 
-#### admin-removed
-
-```
-{
-  "type": "admin-removed"
-  "member": string
-  "clock-value": uint
-}
-```
+#### ADMIN_REMOVED
 
 An admin-removed event is used to drop admin privileges.
-Upon receiving this event a client MUST validate the `chat-id` provided with the updates, MUST ensure that the author of the event is also the target of the event.
+Upon receiving this event a client MUST validate the `chatId` provided with the updates, MUST ensure that the author of the event is also the target of the event.
 
 If the event is valid a client MUST remove the member from the list of `admins` of the chat.
--- a/status-payloads-spec.md
+++ b/status-payloads-spec.md
@@ -179,7 +179,7 @@ The following messages types MUST be supported:
 
 Messages with a `clock` greater than `120` seconds over the whisper timestamp SHOULD be discarded, in order to avoid malicious users to increase the `clock` of a chat arbitrarily.
 
-Messages with a `clock` less than `120` might indicate an attempt to insert messages in the chat history, this is not distinguishable though from a re-transmit from the `datasync` layer. Client MIGHT mark this messages with a warning to the user, or discard them.
+Messages with a `clock` less than `120` might indicate an attempt to insert messages in the chat history, this is not distinguishable though from a re-transmit from the `datasync` layer. A client MAY mark this messages with a warning to the user, or discard them.
 
 `clock` value is used for the message ordering. Due to the used algorithm and distributed nature of the system, we achieve casual ordering which might produce counterintuitive results in some edge cases. For example, when one joins a public chat and sends a message before receiving the exist messages, their message `clock` value might be lower and the message will end up in the past when the historical messages are fetched.
 
--- a/status-payloads-spec.md
+++ b/status-payloads-spec.md
@@ -28,15 +28,15 @@ as various clients created using different technologies.
       - [Message types](#message-types)
       - [Clock vs Timestamp and message ordering](#clock-vs-timestamp-and-message-ordering)
       - [Chats](#chats)
-    - [Contact requests](#contact-requests)
+    - [ContactUpdate](#contact-update)
       - [Payload] (#payload)
-      - [Contact update] (#contact-update)
-      - [Handling contact messages] (#handling-contact-messages)
-    - [SyncInstallation](#sync-installation)
+    - [SyncInstallationContact](#sync-installation-contact)
+      - [Payload](#payload)
+    - [SyncInstallationPublicChat](#sync-installation-public-chat)
       - [Payload](#payload)
     - [PairInstallation](#pair-installation)
       - [Payload](#payload)
-    - [GroupMembershipUpdate](#group-membership-update)
+    - [MembershipUpdateMessage](#membership-update-message)
       - [Payload](#payload)
   - [Upgradability](#upgradability)
   - [Security Considerations](#security-considerations)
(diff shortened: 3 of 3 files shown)


status-group-chats-spec.md
```diff
@@ -10,7 +10,7 @@
 
 - [Abstract](#abstract)
 - [Membership updates](#membership-updates)
-  - [Chat ID](#chat-id)
+  - [Chat ID](#chatId)
```
decanus: should be `#chat-id` so links work

status-group-chats-spec.md
```diff
@@ -28,177 +28,105 @@ This documents describes the group chat protocol used by the status application.
 
 ## Membership updates
 
-Membership updates messages are used to propagate group chat membership changes. The transit format is described in the [Status Payload Specs](status-payload-specs.md). Here we will be describing each specific field.
+Membership updates messages are used to propagate group chat membership changes. The protobuf format is described in the [Status Payload Specs](status-payload-specs.md). Here we will be describing each specific field.
 
-The format is:
+The protobuf messages are:
 
 ```
```
decanus:
```suggestion
```protobuf
```

status-payloads-spec.md
```diff
@@ -65,46 +65,100 @@ If a signature is not present but an author is provided by a layer below, the me
...
-The type `Message` represents a text message exchanged between clients and is identified by the transit tag `c4`.
+The type `ChatMessage` represents a text message exchanged between clients.
 
 #### Payload
 
-Payload is a struct (a compound data type) with the following fields (order is important):
+The protobuf description is:
 
-<!-- TODO: Be more precise in struct description, a la RFC, e.g. TLS style https://tools.ietf.org/html/rfc8446 -->
+```
```
decanus:
```suggestion
```protobuf
```

status-payloads-spec.md
```diff
@@ -134,94 +192,169 @@ All incoming messages can be matched against a chat. Below you can find a table
...
+### Contact Update
 
-### Contact Requests
+`ContactUpdate` is a message exchange to notify peers that either the
+user has been added as a contact, or that information about the sending user have
+changed.
 
-Contact requests consists in 3 kind of messages: `ContactRequest`, `ContactRequestConfirmed` and `ContactUpdate`.
-These messages are used to notify the receiving end that it has been added to the sender's contact. They are identified by the transit tags `c2`, `c3`, `c4` respectively, but they are all interchangeable, meaning a client SHOULD handle them in exactly the same way.  The payload of the 3 messages is identical.
+```
```
decanus:
```suggestion
```protobuf
```

status-payloads-spec.md
```diff
@@ -134,94 +192,169 @@ All incoming messages can be matched against a chat. Below you can find a table
...
 - The profile image is edited
-- A new device has been paired
 
 A client SHOULD also periodically send a `ContactUpdate` to all the contacts, the interval is up to the client, the Status official client sends these updates every 48 hours.
 
+### SyncInstallationContact
+
+`SyncInstallationContact` messages are used to synchronize in a best-effort the contacts to other devices.
+
+```
```
decanus:
```suggestion
```protobuf
```

status-payloads-spec.md
```diff
@@ -65,46 +65,100 @@ If a signature is not present but an author is provided by a layer below, the me
...
+* `TRANSACTION_COMMAND`
 
-These are currently underspecified. We refer to real-world implementations for clients who wish to interoperate.
+##### Sticker content type
 
-<!-- TODO: Ideally specify this, but barring that, link to implementation. -->
+A `ChatMessage` with `STICKER` `Content/Type` MUST also specify the ID of the `Pack` and 
+the `Hash` of the pack, in the `Sticker` field of `ChatMessage`
+
+```
```
decanus:
```suggestion
```protobuf
```

status-payloads-spec.md
```diff
@@ -134,94 +192,169 @@ All incoming messages can be matched against a chat. Below you can find a table
...
+| 4 | ens_name | `string` | ENS name of the contact |
+| 5 | `array[string]` | Array of `system_tags` for the user, this can currently be: `":contact/added", ":contact/blocked", ":contact/request-received"`|
 
-### SyncInstallation
+### SyncInstallationPublicChat
 
-`SyncInstallation` messages are used to synchronize in a best-effort way all the paired installations. It is identified by a transit tag of `p1` 
+`SyncInstallationPublicChat` message is used to synchronize in a best-effort the public chats to other devices.
+
+```
```
decanus:
```suggestion
```protobuf
```

status-payloads-spec.md
```diff
@@ -134,94 +192,169 @@ All incoming messages can be matched against a chat. Below you can find a table
...
-| 3 | chat | `struct {:public? bool :chat-id string}` | A description of a public chat opened by the client |
+| 1 | clock | `uint64` | clock value of the chat | 
+| 2 | id | `string` | id of the chat synced |
 
 ### PairInstallation
 
-`PairInstallation` messages are used to propagate informations about a device to its paired devices. It is identified by a transit tag of `p2` 
+`PairInstallation` messages are used to propagate informations about a device to its paired devices.
+
+```
```
decanus:
```suggestion
```protobuf
```

status-payloads-spec.md
```diff
@@ -134,94 +192,169 @@ All incoming messages can be matched against a chat. Below you can find a table
...
+| 4 | name | `string` | The self-assigned name of the device |
 
-### GroupMembershipUpdate
+### MembershipUpdateMessage and MembershipUpdateEvent
 
-`GroupMembershipUpdate` is a message used to propagate information about group membership changes in a group chat.. It is identified by a transit tag of `g5`.
+`MembershipUpdateEvent` is a message used to propagate information about group membership changes in a group chat.
 The details are in the  [Group chats specs](status-group-chats-spec.md)
 
+```
```
decanus:
```suggestion
```protobuf
```

decanus (changes requested):

status-group-chats-spec.md
```diff
@@ -28,177 +28,105 @@ This documents describes the group chat protocol used by the status application.
...
-         ["b", "b-value"],
-         ["clock-value", 1]
-       ]
-     ],
-     "chat-id"
-   ]
-```
-
-This structure is then stringified collapsing all whitespaces and the `Keccak256` of the string is then signed using its private key by the author and added to the payload.
+The signature for each event is calculated by encoding each `MembershipUpdateEvent` in its protobuf representation
```
arnetheduck: protobuf is a non-deterministic format - there are many valid serializations for the same data - this will not fly.

arnetheduck: fwiw,  here's the relevant documentation link: https://developers.google.com/protocol-buffers/docs/encoding#order

status-payloads-spec.md
```diff
@@ -134,94 +192,169 @@ All incoming messages can be matched against a chat. Below you can find a table
...
 #### Payload
 
 | Field | Name | Type | Description |
 | ----- | ---- | ---- | ---- |
-| 1| installation-id | `string` | A randomly generated id that identifies this device |
-| 2 | device-type | `string` | The OS of the device `ios`,`android` or `desktop` |
-| 3 | name | `string` | The self-assigned name of the device |
-| 4 | fcm-token | `string` | The FCM Token used by mobile devices for push notifications |
+| 1 | clock | `uint64` | clock value of the chat | 
+| 2| installation_id | `string` | A randomly generated id that identifies this device |
```
arnetheduck, line 292:
```suggestion
| 2 | installation_id | `string` | A randomly generated id that identifies this device |
```

status-group-chats-spec.md
cammellos: I don't think that's a problem, I believe we had already a discussion about this at some point.

Basically the fact that protobuf does not guarantees a deterministic format (across versions I believe, same implementation generally is deterministic, but let's assume as well it isn't), does not impact this, for example:

1) I create an event, I encode it to protobuf and sign the bytes generated, send over `signature+bytes`.
2) A client receives it, verify `signature` against `bytes`, decodes the message. 

At no point a client should re-encode the message nor the message should be re-encoded (as well because that's likely to break in protobuf, as unknown fields are simply ignored, so it would break the protocol at the first upgrade), if the message is to be relayed, the original `signature+bytes` should be used, unchanged.

The event itself when it comes to the signature, is treated as just as bytes, the fact that generating those bytes is not deterministic as no impact on the validity of the signature.

If client A (for whatever reason, even if currently there's no use case for this, as re-transmitting uses the raw payload), re-encodes the message twice, and gets two different results, it is also not a problem as the protocol is idempotent so no change will result from duplicates, even if they are not recognized as such.

What do you think?

status-group-chats-spec.md
```diff
@@ -28,177 +28,105 @@ This documents describes the group chat protocol used by the status application.
...
+  EventType type = 4;
+
+  enum EventType {
+    UNKNOWN = 0;
+    CHAT_CREATED = 1;
+    NAME_CHANGED = 2;
+    MEMBERS_ADDED = 3;
+    MEMBER_JOINED = 4;
+    MEMBER_REMOVED = 5;
+    ADMINS_ADDED = 6;
```
arnetheduck: how does the protocol establish a total order of these events?

cammellos: We use clock value (lamport timestamp essentially), ties broken by message-id , same as chat messages essentially

status-group-chats-spec.md
arnetheduck: I think this is fishy because it becomes possible to sign byte sequences whose deserialized representation is unexpected - you're signing some random bytes at that point and not a particular message as it logically appears to the receiver - you're proving that you have the private key, but not that it was your intent to sign a particular instruction, which seems to open up the signature for interpretation as well, weakening the value of it.

I agree it's mitigated by the custom (non-protobuf) serialization format in `events` and that the use case is somewhat casual (chat). I anticipate surprises as it starts getting abused.

> unknown fields are simply ignored

not sure what you mean here - it's a feature of protobuf to be able to reencode and alter a message without understanding all the fields - that includes preserving unknown fields (perhhaps not all protobuf implementations do this currently, c++/python did last I checked)

decanus: What we did in MVDS is order the fields to sign, see `message indentifiers`. https://specs.vac.dev/mvds.html#payloads

cammellos: Not sure I fully understand your point, 
you are not signing some "random bytes", you are signing the encode version of the message you are about to send.
It does imply that you trust the protobuf library to encode the message non maliciously, is that your concern? (that goes to any library that you use to generate the message though, whether you are concatenating the field etc, regardless of the signature scheme)

The method used is the one suggested by the primary author of protobuf https://groups.google.com/d/msg/protobuf/iCA5ALRtDUI/vUiAl6GQqUAJ &  https://capnproto.org/ .

The alternative so far proposed (to sign each field, concatenating etc), are fine with static protocol that don't change often, that's why we used it in the first place in this one, but are more prone to break as newly added field might be ignored by older clients, so won't be used to calculate the signature.

Unkowns fields were ignored in proto 3, seems like they corrected it in proto 3.5 (we still use 3 ourselves):

https://developers.google.com/protocol-buffers/docs/proto3#unknowns

So the method used before (concatenating each field separately), would break when adding a new field (or we have to have a new field with the new signature scheme, maintaining the old one).

What method are you proposing exactly as an alternative?

MVDS strategy for example, given we use protobuf3 , would break if we ever added a new field to the datasync protobuf message, and the choice would be, either to exclude it from messageID calculation (which might be fine with messageID maybe, definitely not with signatures), or account that with backward compatible code.

cammellos:
@decanus updated according the feedback, could you take another look please?

decanus:
@cammellos I need to reconsider the signature part, will get back to you asap.

decanus:
Also, @cammellos protobuf version should be specified somewhere.

status-group-chats-spec.md
```diff
@@ -28,177 +28,105 @@ This documents describes the group chat protocol used by the status application.
...
+  }
 }
 ```
 
 ### Chat ID
 
-Each membership update MUST be sent with a corresponding `chat-id`. 
-The format of this chat id MUST be a string, [UUID](https://tools.ietf.org/html/rfc4122 ), concatenated with the hex-encoded public key of the creator of the chat. This chat-id MUST be validated by all clients, and MUST be discarded if it does not follow these rules.
+Each membership update MUST be sent with a corresponding `chatId`. 
+The format of this chat id MUST be a string, [UUID](https://tools.ietf.org/html/rfc4122 ), concatenated with the hex-encoded public key of the creator of the chat. This chatId MUST be validated by all clients, and MUST be discarded if it does not follow these rules.
```
adambabik:
```suggestion
The format of this chat ID MUST be a string of [UUID](https://tools.ietf.org/html/rfc4122) concatenated with the hex-encoded public key of the creator of the chat. This chat ID MUST be validated by all clients, and MUST be discarded if it does not follow these rules.
```

Do we use `-` for concatenation or nothing?

status-payloads-spec.md
```diff
@@ -114,15 +168,19 @@ Message types are required to decide how a particular message is encrypted and w
...
+* `ONE_TO_ONE` is a message to the public group
+* `PUBLIC_GROUP` is a private message
+* `PRIVATE_GROUP` is a message to the private group.
 
 #### Clock vs Timestamp and message ordering
 
 `timestamp` MUST be Unix time calculated when the message is created in milliseconds. This field SHOULD not be relied upon for message ordering.
 
-`clock` SHOULD be calculated using the algorithm of [Lamport timestamps](https://en.wikipedia.org/wiki/Lamport_timestamps). When there are messages available in a chat, `clock`'s value is calculated based on the last received message in a particular chat: `last-message-clock-value + 1`. If there are no messages, `clock` is initialized with `timestamp * 100`'s value.
+`clock` SHOULD be calculated using the algorithm of [Lamport timestamps](https://en.wikipedia.org/wiki/Lamport_timestamps). When there are messages available in a chat, `clock`'s value is calculated based on the last received message in a particular chat: `max(timeNowInMs, last-message-clock-value + 1)`. If there are no messages, `clock` is initialized with `timestamp * 100`'s value.
```
adambabik: Why `* 100`?

adambabik (approved):

status-group-chats-spec.md
```diff
@@ -28,177 +28,105 @@ This documents describes the group chat protocol used by the status application.
...
 ## Membership updates
 
-Membership updates messages are used to propagate group chat membership changes. The transit format is described in the [Status Payload Specs](status-payload-specs.md). Here we will be describing each specific field.
-
-The format is:
+Membership updates messages are used to propagate group chat membership changes. The protobuf format is described in the [Status Payload Specs](status-payload-specs.md). Here we will be describing each specific field.
+
+The protobuf messages are:
+
+```protobuf
```
oskarth, line 35: 1. Could we specify protobuf version used in these fragments? Either globally or per section

2. If we have a package name, can we specify these? https://developers.google.com/protocol-buffers/docs/proto If we don't, could we consider adding it?

status-group-chats-spec.md
```diff
@@ -28,177 +28,105 @@ This documents describes the group chat protocol used by the status application.
...
+// MembershipUpdateMessage is a message used to propagate information
+// about group membership changes.
+message MembershipUpdateMessage {
+  // The chat id of the private group chat
+  string chat_id = 1;
+  // A list of events for this group chat, first x bytes are the signature, then is a
```
oskarth: What is x?

status-group-chats-spec.md
```diff
@@ -28,177 +28,105 @@ This documents describes the group chat protocol used by the status application.
...
-The format is:
+Membership updates messages are used to propagate group chat membership changes. The protobuf format is described in the [Status Payload Specs](status-payload-specs.md). Here we will be describing each specific field.
+
+The protobuf messages are:
+
+```protobuf
+// MembershipUpdateMessage is a message used to propagate information
+// about group membership changes.
+message MembershipUpdateMessage {
+  // The chat id of the private group chat
```
oskarth, line 39: Is this unique to private group chats? vs "group chat" as a construct (I know we don't - currently, anymore - support public group chats).

status-group-chats-spec.md
```diff
@@ -28,177 +28,105 @@ This documents describes the group chat protocol used by the status application.
...
+  ChatMessage message = 3;
+}
 
-```
-{
-  "events": [struct {"type": string, "member": string, "members": [string], "clock-value": uint, "name": string],
-  "signature": string,
-  "chat-id": string
+message MembershipUpdateEvent {
+  // Lamport timestamp of the event
```
oskarth: can we put a reference to where in the specs this is elaborated on?

status-group-chats-spec.md
```diff
@@ -28,177 +28,105 @@ This documents describes the group chat protocol used by the status application.
...
+  uint64 clock = 1;
+  // List of public keys of objects of the action
```
oskarth: can we rephrase this? "of objects" seems unclear to me

status-group-chats-spec.md
```diff
@@ -28,177 +28,105 @@ This documents describes the group chat protocol used by the status application.
...
+  repeated string members = 2;
+  // Name of the chat for the CHAT_CREATED/NAME_CHANGED event types
+  string name = 3;
+  // The type of the event
+  EventType type = 4;
+
+  enum EventType {
```
oskarth, line 58: a reference to semantics and guarantees in different part of spec would be useful

status-payloads-spec.md
```diff
@@ -65,46 +65,100 @@ If a signature is not present but an author is provided by a layer below, the me
 
 ## Encoding
 
-The payload is encoded using [Transit format](https://github.com/cognitect/transit-format). Transit was chosen over JSON in order to reduce the bandwidth.
+The payload is encoded using [Protobuf](https://developers.google.com/protocol-buffers)
```
oskarth, line 68: specify version

status-payloads-spec.md
```diff
@@ -65,46 +65,100 @@ If a signature is not present but an author is provided by a layer below, the me
...
 
 ## Types of messages
 
 ### Message
 
-The type `Message` represents a text message exchanged between clients and is identified by the transit tag `c4`.
+The type `ChatMessage` represents a text message exchanged between clients.
```
oskarth: text -> chat? text to me seems to imply plain text

status-payloads-spec.md
```diff
@@ -65,46 +65,100 @@ If a signature is not present but an author is provided by a layer below, the me
...
 
-Payload is a struct (a compound data type) with the following fields (order is important):
+The protobuf description is:
 
-<!-- TODO: Be more precise in struct description, a la RFC, e.g. TLS style https://tools.ietf.org/html/rfc8446 -->
+```protobuf
+message ChatMessage {
+  // Lamport timestamp of the chat message
+  uint64 clock = 1;
+  // Unix timestamps in milliseconds, currently not used as we use whisper as more reliable, but here
```
oskarth: I think I know what you mean, but this can probably be rephrased for clarity

status-payloads-spec.md
```diff
@@ -65,46 +65,100 @@ If a signature is not present but an author is provided by a layer below, the me
...
-| 1 | text | `string` | The text version of the message content |
-| 2 | content type | `enum` (more in [Content types](#content-types)) | See details |
-| 3 | message type | `enum` (more in [Message types](#message-types)) | See details |
-| 4 | clock | `int64` | See details |
-| 5 | timestamp | `int64` | See details |
-| 6 | content | `struct { chat-id string, text string, response-to string }` | The chat-id of the chat this message is destined to, the text of the content and optionally the id of the message it is responding to|
+| 1 | clock | `uint64` | The clock of the chat|
+| 2 | timestamp | `uint64` | The sender timestamp at message creation |
+| 3 | text | `string` | The content of the message |
+| 4 | response_to | `string` | The ID of the message replied to |
```
oskarth, line 130: would it make sense to have an ID type with restrictions to it? even if it is something like `stringid` and then elaborated on above

status-payloads-spec.md
```diff
@@ -65,46 +65,100 @@ If a signature is not present but an author is provided by a layer below, the me
...
+| 8 | content_type | `ContentType` | The type of the content of the message | 
+| 9 | payload | `Sticker|nil` | The payload of the message based on the content type |
 
 #### Content types
 
 Content types are required for a proper interpretation of incoming messages. Not each message is plain text but may carry a different information.
 
 The following content types MUST be supported:
-* `text/plain` identifies a message which content is a plain text.
+* `TEXT_PLAIN` identifies a message which content is a plain text.
```
oskarth:
```suggestion
* `TEXT_PLAIN` identifies a message which content is plaintext.
```

status-payloads-spec.md
```diff
@@ -134,94 +192,169 @@ All incoming messages can be matched against a chat. Below you can find a table
...
 
-### GroupMembershipUpdate
+### MembershipUpdateMessage and MembershipUpdateEvent
 
-`GroupMembershipUpdate` is a message used to propagate information about group membership changes in a group chat.. It is identified by a transit tag of `g5`.
+`MembershipUpdateEvent` is a message used to propagate information about group membership changes in a group chat.
 The details are in the  [Group chats specs](status-group-chats-spec.md)
 
+```protobuf
+message MembershipUpdateEvent {
```
oskarth: I feel like I have seen this before, is there a reason to duplicate it?

status-payloads-spec.md
```diff
@@ -134,94 +192,169 @@ All incoming messages can be matched against a chat. Below you can find a table
...
+| 3 | message | `ChatMessage` | An optional message, described in [Message](#message) |
 
 ## Upgradability
 
 There are two ways to upgrade the protocol without breaking compatibility:
 
-- Struct fields can be enriched with a new key, which will be ignored by old clients.
-- An element can be appended to the `Transit` array, which will also be ignored by old clients.
+- Accretion is always supported
+- Deletion of existing is not supported and might break compatibility
```
oskarth:
```suggestion
- Deletion of existing fields or messages is not supported and might break compatibility
```

status-payloads-spec.md
```diff
@@ -134,94 +192,169 @@ All incoming messages can be matched against a chat. Below you can find a table
...
 
 ## Security Considerations
 
 TBD.
 
 ## Design rationale
```
oskarth, line 312: so now we don't even have a rationale! pff ;)

oskarth (changes requested):
Awesome work on moving everything over to protobuf!

Initial comments

status-group-chats-spec.md
cammellos: we do, updated

status-payloads-spec.md
cammellos: that's actually not the case anymore, we multiplied by `100` to separate the timestamp from the  `sequence` number of the message, but it's not really needed, so we just use a timestamp now, updated

status-group-chats-spec.md
cammellos: I will add the version, currently we use `protobuf` as a package name (in go it does not have change the generated code, so we can pick anything), what should we go for? (I might update the specs though in a separate PR, once the change is made)

status-group-chats-spec.md
cammellos: 65 bytes, added

status-group-chats-spec.md
cammellos: Yes, that's only used for private group chats

status-payloads-spec.md
cammellos: Not sure it's actually even worth mentioning is not used, the only way we use it is  cosmetic (to show a user what time a message has been sent, but it's purely cosmetic (another client might chose not to do so), what do you think?

status-payloads-spec.md
cammellos: I would probably keep it a string, because technically it is a string (so strings need to be used in protobuf), and then is validated (in the same way for other fields), I have always taken this as what is currently sent across the wire, but never really spent time questioning it.

Essentially, do we want this field to represent what is being sent in `protobuf` or we want it to diverge from that in the specs so that it's easier to interpret for someone reading it?

I don't have a strong opinion myself, either are fine

status-payloads-spec.md
cammellos: Moved to `status-group-chats-spec.md`

status-group-chats-spec.md
cammellos: the best I could come up with is `of the targets`, not sure is much clearer though :)

cammellos:
@oskarth addressed feedback and left a few questions, I have specified version of `protobuf` in `status-payload-specs.md` (saying that is `proto3` unless specified otherwise), as we use `protobuf` throughout the whole stack

status-group-chats-spec.md
arnetheduck: Assuming that the reading client parses the data and then displays it to the user as "signed", there could be question about what exactly was signed: was it the parsed message or the byte-stream? I wonder then if the sender can construct a message, sign it and then claim that what they signed was not what the recipient understood it to be (after parsing).

You make a good point that if idempotency is not needed, the serialization matters less - in MVDS the message id is important for the correctness of the sync - if that is not the case here, it's less of a problem than I initially assumed. Thanks.

status-group-chats-spec.md
decanus: Yeah we should defo specify, in MVDS we also specify which version we use.

decanus: @cammellos @oskarth you can add `syntax = "proto3";` to the protocol buffer itself as the first line before the message.

cammellos: just to clarify, we do always specify both `syntax` (`proto3`) and package (`protobuf`, the package name we use is just `protobuf`), just wondering whether we should use a more meaningful package name

status-group-chats-spec.md
cammellos: I don't think that would be possible, essentially non-repudiation and integrity still stand, and the result is always unambiguous (we never sign the parsed message but always the byte-stream), even if there's  more than one way to serialize the message (due to the order of fields not being specified/enforced essentially, given `protobuf` is map).

There can be no doubt on the origin and the integrity of the message, whether there might be some doubt on the interpretation it would be a problem in the protocol itself (say for example the two clients have mismatched protobuf specifications).
In general protobuf being not self-describing might lead to such issues, but they would come up regardless of the signature scheme used.

thanks for the feedback

status-client-spec.md
```diff
@@ -67,6 +67,10 @@ Implementing a Status clients means implementing the following layers. Additiona
 | Transport privacy | Routing, Metadata protection    | Whisper                      |
 | P2P Overlay       | Overlay routing, NAT traversal  | devp2p                       |
 
+### Protobuf
+
+We use `protobuf` in different layers, the version used is `proto3` unless stated otherwise.
```
decanus:
```suggestion
We use [`protobuf`](https://developers.google.com/protocol-buffers/) in different layers, the version used is `proto3` unless stated otherwise.
```

status-group-chats-spec.md
```diff
@@ -28,177 +28,123 @@ This documents describes the group chat protocol used by the status application.
...
+
+The protobuf messages are:
+
+```protobuf
+// MembershipUpdateMessage is a message used to propagate information
+// about group membership changes.
+message MembershipUpdateMessage {
+  // The chat id of the private group chat
+  string chat_id = 1;
+  // A list of events for this group chat, first 65 bytes are the signature, then is a
```
decanus, line 41: why do this and not just create a new internal message called event

```
message Event {
 repeated bytes signature = 1;
 MembershipUpdateEvent update = 2;
}
```

Not criticial but worth knowing the rationale.

status-payloads-spec.md
```diff
@@ -114,15 +167,19 @@ Message types are required to decide how a particular message is encrypted and w
...
+`clock` SHOULD be calculated using the algorithm of [Lamport timestamps](https://en.wikipedia.org/wiki/Lamport_timestamps). When there are messages available in a chat, `clock`'s value is calculated based on the last received message in a particular chat: `max(timeNowInMs, last-message-clock-value + 1)`. If there are no messages, `clock` is initialized with `timestamp`'s value.
+
+Messages with a `clock` greater than `120` seconds over the whisper timestamp SHOULD be discarded, in order to avoid malicious users to increase the `clock` of a chat arbitrarily.
+
+Messages with a `clock` less than `120` might indicate an attempt to insert messages in the chat history, this is not distinguishable though from a re-transmit from the `datasync` layer. Client MIGHT mark this messages with a warning to the user, or discard them.
```
decanus:
```suggestion
Messages with a `clock` less than `120` might indicate an attempt to insert messages in the chat history, this is not distinguishable though from a re-transmit from the `datasync` layer. A client MAY mark this messages with a warning to the user, or discard them.
```

MIGHT is not in rfc2119

decanus (changes requested):

status-group-chats-spec.md
cammellos: We sign the raw bytes, so those need to be sent over and easily extracted.
So your suggestion would be:

```
message Event {
 repeated bytes signature = 1;
 repeated bytes payload = 2; // This is actually a MembershipUpdateEvent
}
```

which works fine, and it is more explicit.

I slightly prefer the concatenation mainly for ease of use, clients will take the event and store it as it, unchanged, so it can be relayed to third party, and probably save a few bytes.

Either though are perfectly fine I think.

decanus: @cammellos ok that makes sense, lets keep it as is then.

status-group-chats-spec.md
oskarth: What about `status.messaging`? Looking at https://github.com/status-im/status-go/tree/baa0767c263fa25fdd8d90a42d13f1f1fffce804/protocol/protobuf it seems like there are 3 short-ish protobufs, so having one namespace for all of them seems good enough for them. Once we do different domains, we can have `status.foo`, and we'll also have `vac.waku` etc.

status-payloads-spec.md
oskarth: Keeping it as a string here makes sense for accuracy

One idea would be to have a small section below on restriction/validation to explain additional requirements we have

status-group-chats-spec.md
oskarth: This question seems to come up quite frequently, it might be a good idea to add it to a rationale/FAQ section at the end so we don't end up having the same conversation over and over again :)

oskarth (approved):
Thanks! Left a comment re namespace and some other stuff, otherwise lgtm https://github.com/status-im/specs/pull/80/files#r369364189

PR #87, opened 2020-03-24
Move Status EIPs to stable

oskarth:
This PR moves EIPs spec from draft to stable. This means it is an accurate description of what is currently in production for the v1 app.

Changes to this spec can still occur but they should be cocsmetic ones, errata and clarifications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,9 @@ Every spec has its own lifecycle that shows its maturity. We indicate this in a 
 
 ### Stable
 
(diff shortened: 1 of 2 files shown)


oskarth:
> Register at push notification system

> If you want to receive and send push notifications, you MUST register a push notification server. This part is currently underspecified. You MAY choose to not do this.

I believe this has been removed in v1? Please confirm. E.g. @yenda

3esmit (approved):

PR #88, opened 2020-03-24
Move Whisper Mailserver Usage to stable

oskarth:
This PR moves mailserver spec from draft to stable. This means it is an accurate description of what is currently in production for the v1 app.

Changes to this spec can still occur but they should be cosmetic ones, errata and clarifications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,10 @@ The following SIPs are stable and running in production.
 
 - [Status Secure Transport Specification](status-secure-transport-spec.md). How Status provide a secure transport with conversational security properties.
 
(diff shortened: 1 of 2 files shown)


cammellos (approved):
Overall is fine, the only thing I am not so sure about is:

`Effectively, it knows which topics the node is interested in`

That's not accurate I believe, as currently (for whisper, not waku, but I take this is whisper only), we send a bloom filter, not plain topics, therefore it has no more knowledge then any other node the user is connected to.

I would change to something like:
`Effectively, it will have access to the bloom filter of topics that the user is interested in`

adambabik:
> Currently Status Gmbh provides mailservers in an altruistic manner,

Is it still Status Gmbh?

adambabik (approved):

oskarth:
Thanks! Addressed both

PR #92, opened 2020-03-24
Move payloads to stable

oskarth:
This PR moves payloads spec from draft to stable. This means it is an accurate description of what is currently in production for the v1 app.

Changes to this spec can still occur but they should be cosmetic ones, errata and clarifications.

### DIFF
--- a/status-payloads-spec.md
+++ b/status-payloads-spec.md
@@ -1,8 +1,10 @@
 # Status Message Payloads Specification
 
-> Version: 0.1 (Draft)
+> Version: 0.2
 >
-> Authors: Adam Babik <adam@status.im>, Oskar Thorén <oskar@status.im> (alphabetical order)
+> Status: Stable
+>
+> Authors: Adam Babik <adam@status.im>, Andrea Maria Piana <andreap@status.im>, Oskar Thorén <oskar@status.im> (alphabetical order)
 
 ## Abstract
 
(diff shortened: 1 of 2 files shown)


corpetty (changes requested):
questions/comments/suggestions before my approval
- L82-128 - do we need this with the next section? I updated it to latest `status-go` protocol commit, but a reference may be better suited if we define it in a table directly afterwards.
- L148-155 - is this list still complete as of V1.2? @adambabik @cammellos 
- L167 - complete other `Content/Type`s: `STATUS`, `EMOJI`, and `TRANSACTION_COMMAND` definitions
- L173 - address TODO
- L177-178 - switch definition of these two, they're backwards
- L191 - `exist` -> `existing`

status-payloads-spec.md
```diff
@@ -81,15 +84,19 @@ The protobuf description is:
 message ChatMessage {
```
oskarth, line 84: @corpetty 

Do you mean this payload section? Why wouldn't we?

L82-128 - do we need this with the next section? I updated it to latest status-go protocol commit, but a reference may be better suited if we define it in a table directly afterwards.

oskarth:
@corpetty Thanks for the feedback! Could you please do the review in-line? The line numbers are quite confusing and they also change. It is much easier to have a conversation and do suggestions etc in a specific locale. E.g. I noticed you pushed a commit with fixes, but I have no idea what that fixes of your comments.

I know the file diff is a bit off, so you gotta expand to see lines not changed. Even that doesn't work all the time. I suggest pushing a commit with the comment in-line at the appropriate place, then we can comment and discuss on it in-line. WDYT?

PR #94, opened 2020-03-26
feature/remark

decanus:


package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@vacp2p/specs",
```
oskarth: Is this correct?

decanus: fixed, copy pasta ninja

oskarth (approved):
assuming it works

PR #101, opened 2020-04-14
update whisper key derivation path

gravityblast:
@corpetty @oskarth does it make sense to update the derivation path in the `Terms Glossary` as well?

### DIFF
--- a/docs/stable/2-account.md
+++ b/docs/stable/2-account.md
@@ -111,7 +111,7 @@ Everything else associated with the contact is either verified or derived from t
 | ---- | ----------- |
 | privkey | ECDSA secp256k1 private key |
 | pubkey | ECDSA secp256k1 public key |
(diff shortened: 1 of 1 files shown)


decanus:
I'd say yes

andremedeiros (approved):

gravityblast:
I don't have permissions to merge, feel free to merge it when you want

oskarth:
@gravityblast Could you elaborate on this change a bit? Was the previous spec wrong? Is this a change? In which version?

@decanus lets not merge so quickly next time :)

gravityblast:
@oskarth yes the spec was wrong. The `Terms Glossary` said that the whisper key uses this derivation path `m/44'/60'/0'/0/0`, which was the one used before the multi-account implementation.
This PR updates that path with the one we are currently using: `m/43'/60'/1581'/0'/0`.

PR #107, opened 2020-04-30
Image content type for chat messages

cammellos:
Adds a content type for image messages.

For now the supported formats are `JPEG`, `PNG`, `GIF` and `WEBP `, which should be enough to cover most cases.

Ignore the first commit , as it only copies the stable specs to draft.
https://github.com/status-im/specs/pull/107/commits/ca530d9d80681e9d6442fd10cac3e0f2607d88b7

### DIFF
--- a/docs/draft/6-payloads.md
+++ b/docs/draft/6-payloads.md
@@ -1,0 +1,335 @@
+---
+permalink: /spec/6
+parent: Stable specs
+title: 6/PAYLOADS
+---
+
+# 6/PAYLOADS
+
+> Version: 0.2
+>
+> Status: Stable
+>
+> Authors: Adam Babik <adam@status.im>, Andrea Maria Piana <andreap@status.im>, Oskar Thorén <oskar@status.im> (alphabetical order)
+
+## Abstract
+
+This specifications decribes how the payload of each message in Status looks
+like. It is primarly centered around chat and chat-related use cases.
+
+The payloads aims be flexible enough to support messaging but also cases
+described in the [Status Whitepaper](https://status.im/whitepaper.pdf) as well
+as various clients created using different technologies.
+
+## Table of Contents
+
+- [Status Message Payloads Specification](#status-message-payloads-specification)
+  - [Abstract](#abstract)
+  - [Table of Contents](#table-of-contents)
+  - [Introduction](#introduction)
+  - [Payload wrapper](#payload-wrapper)
+  - [Encoding](#encoding)
+  - [Types of messages](#types-of-messages)
+    - [Message](#message)
+      - [Payload](#payload)
+      - [Payload](#payload-1)
+      - [Content types](#content-types)
+        - [Sticker content type](#sticker-content-type)
+      - [Message types](#message-types)
+      - [Clock vs Timestamp and message ordering](#clock-vs-timestamp-and-message-ordering)
+      - [Chats](#chats)
+    - [Contact Update](#contact-update)
+      - [Payload](#payload-2)
+      - [Contact update](#contact-update-1)
+    - [SyncInstallationContact](#syncinstallationcontact)
+      - [Payload](#payload-3)
+    - [SyncInstallationPublicChat](#syncinstallationpublicchat)
+      - [Payload](#payload-4)
+    - [PairInstallation](#pairinstallation)
+      - [Payload](#payload-5)
+    - [MembershipUpdateMessage and MembershipUpdateEvent](#membershipupdatemessage-and-membershipupdateevent)
+  - [Upgradability](#upgradability)
+  - [Security Considerations](#security-considerations)
+  - [Design rationale](#design-rationale)
+
+## Introduction
+
+In this document we describe the payload format and some special considerations.
+
+## Payload wrapper
+
+All payloads are wrapped in a [protobuf record](https://developers.google.com/protocol-buffers/)
+record:
+
+```protobuf
+message StatusProtocolMessage {
+  bytes signature = 4001;
+  bytes payload = 4002;
+}
+```
+
+`signature` is the bytes of the signed `SHA3-256` of the payload, signed with the key of the author of the message.
+The signature is needed to validate authorship of the message, so that the message can be relayed to third parties.
+If a signature is not present but an author is provided by a layer below, the message is not to be relayed to third parties and it is considered plausibly deniable.
+
+## Encoding
+
+The payload is encoded using [Protobuf](https://developers.google.com/protocol-buffers)
+
+## Types of messages
+
+### Message
+
+The type `ChatMessage` represents a chat message exchanged between clients.
+
+#### Payload
+
+The protobuf description is:
+
+```protobuf
+message ChatMessage {
+  // Lamport timestamp of the chat message
+  uint64 clock = 1;
+  // Unix timestamps in milliseconds, currently not used as we use whisper as more reliable, but here
+  // so that we don't rely on it
+  uint64 timestamp = 2;
+  // Text of the message
+  string text = 3;
+  // Id of the message that we are replying to
+  string response_to = 4;
+  // Ens name of the sender
+  string ens_name = 5;
+  // Chat id, this field is symmetric for public-chats and private group chats,
+  // but asymmetric in case of one-to-ones, as the sender will use the chat-id
+  // of the received, while the receiver will use the chat-id of the sender.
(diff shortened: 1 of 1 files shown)


docs/draft/6-payloads.md
```diff
@@ -0,0 +1,359 @@
...
+  // Text of the message
+  string text = 3;
+  // Id of the message that we are replying to
+  string response_to = 4;
+  // Ens name of the sender
+  string ens_name = 5;
+  // Chat id, this field is symmetric for public-chats and private group chats,
+  // but asymmetric in case of one-to-ones, as the sender will use the chat-id
+  // of the received, while the receiver will use the chat-id of the sender.
+  // Probably should be the concatenation of sender-pk & receiver-pk in alphabetical order
```
corpetty: need this commentary?

cammellos: I think we can remove it

docs/draft/6-payloads.md
```diff
@@ -0,0 +1,359 @@
...
+| 7 | message_type | `MessageType` | The type of message, different for one-to-one, public or group chats |
+| 8 | content_type | `ContentType` | The type of the content of the message | 
+| 9 | payload | `Sticker|Image|nil` | The payload of the message based on the content type |
+
+#### Content types
+
+Content types are required for a proper interpretation of incoming messages. Not each message is plain text but may carry a different information.
+
+The following content types MUST be supported:
+* `TEXT_PLAIN` identifies a message which content is a plaintext.
```
jakubgs, line 156: Does `TEXT_PLAIN` include markdown, or `(is there|should there be)` a `TEXT_MARKDOWN`?

docs/draft/6-payloads.md
```diff
@@ -0,0 +1,359 @@
...
+  bytes payload = 1;
+  ImageType type = 2;
+}
+
+enum ImageType {
+  UNKNOWN_IMAGE_TYPE = 0;
+  IMAGE_JPEG = 1;
+  IMAGE_PNG = 2;
+  IMAGE_WEBP = 3;
+}
```
andremedeiros: `IMAGE_GIF`?

andremedeiros:
Would be good to have `payload` limit documented + address multipart (for large images)

docs/draft/6-payloads.md
cammellos: It includes markdown, yes, we could rename it possibly

jakubgs: I think it would make more sense that way. When I'm implementing a client and I see `TEXT_PLAIN` I don't expect it to spew at me Markdown. The `PLAIN` part being key here.

jakubgs: Supporting both should make most sense. Or making `TEXT_PLAIN` mandatory and `TEXT_MARKDOWN` optional.

cammellos: Not sure we want to send them both over the wire, having a textual representation makes sense (say for the chat preview), but then you get into issues about consistency (say `TEXT_MARKDOWN` is different from `TEXT_PLAIN` for malicious reasons).

So just sending over markdown and then parsing and getting a textual representation seems a bit safer, but has the drawback that is harder to gracefully introduce new markdown tags, although is not trivial in the other instance either.

For now I think we should just change `TEXT_PLAIN` to `TEXT_MARKDOWN` and handle textual representation on the receivers end.

jakubgs: Fair.

docs/draft/6-payloads.md
```diff
@@ -0,0 +1,359 @@
+---
+permalink: /spec/6
+parent: Draft specs
+title: 6/PAYLOADS
```
oskarth, line 4: Is this meant to replace the existing 6/PAYLOADS or augment it?

oskarth: I'm not sure, but it might be worth making this a more minimal "payload-image" draft spec. Then 6/payloads could be augmented with a MAY support <image payload spec>.

Right now it is quite hard to tell what has changed, and the PR isn't mergable as is since we already have 6 marked as stable. (EDIT: Going through manual last commit and view changes makes diff clearer)

docs/draft/6-payloads.md
oskarth: Is this part of the PR/change? Seems out of scope AFAICT.

docs/draft/6-payloads.md
```diff
@@ -0,0 +1,359 @@
...
+```
+
+##### Image content type
+
+A `ChatMessage` with `IMAGE` `Content/Type` MUST also specify the `payload` of the image
+and the `type`
+
+```protobuf
+message ImageMessage {
+  bytes payload = 1;
```
oskarth, line 184: Does this assume an image fits into one message?

docs/draft/6-payloads.md
```diff
@@ -0,0 +1,359 @@
...
+The details are in the  [Group chats specs](status-group-chats-spec.md)
+
+## Upgradability
+
+There are two ways to upgrade the protocol without breaking compatibility:
+
+- Accretion is always supported
+- Deletion of existing fields or messages is not supported and might break compatibility
+
+## Security Considerations
```
oskarth, line 351: Anything specific to keep in mind when loading images/gifs?

cc @corpetty

oskarth (changes requested):

docs/draft/6-payloads.md
cammellos: Yes, 
if fits in one `protobuf` message.
The idea is that any multipart is not done at this level (i.e not at the application layer), but at a lower level, either before waku/encryption (that's to be decided). If the message is too big, the whole protobuf payload is chunked, so any application message that exceeds the maximum size can be chunked (although we haven't really talked about multipart as of now)

docs/draft/6-payloads.md
```diff
@@ -0,0 +1,359 @@
...
+  // of the received, while the receiver will use the chat-id of the sender.
+  // Probably should be the concatenation of sender-pk & receiver-pk in alphabetical order
+  string chat_id = 6;
+
+  // The type of message (public/one-to-one/private-group-chat)
+  MessageType message_type = 7;
+  // The type of the content of the message
+  ContentType content_type = 8;
+
+  oneof payload {
```
oskarth, line 111: Do we specify message size somewhere? Might be good to do it explicit in payload / image spec, e..g Status supports messages up to <X size>

cammellos: Currently it's a bit tricky, we can set a limit, but there's no guarantee that it will get through, as each whisper node sets its own max message limit. 
But we can suggest a safe limit, but we need to do a bit more exploratory work I think, to check image sizes , pow etc

docs/draft/6-payloads.md
```diff
@@ -0,0 +1,359 @@
...
+There are two ways to upgrade the protocol without breaking compatibility:
+
+- Accretion is always supported
+- Deletion of existing fields or messages is not supported and might break compatibility
+
+## Security Considerations
+
+-
+
+## Design rationale
```
oskarth, line 355: ## Copyright

Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

(realized this was missing in original, an oversight - possibly separate pr).

docs/draft/6-payloads.md
oskarth: Thoughts on this @cammellos?

cammellos: I don't have any opinion on this, operationally it's simpler to copy specs to draft, move, and copy over when it's stable. Visibility can be ensured by always copy a new DRAFT once a new STABLE is added (that can be a simple hook), but that also comes with some issue, for example if you partially implement a draft.
Not sure is worth though polishing the process so much as generally speaking is fairly static and we make few changes. So anything is good with me really.

cammellos:
> Would be good to have `payload` limit documented + address multipart (for large images)

Not sure, I would not add multipart as this is not part of image sending or payload in general, we haven't really settled on whether it will be built and which layer it would be taken care of (my gut feelings is that it will be at a much lower layer). 
Similarly `payload` has not limit, the limit is dictated by the transport (which should provide multipart if necessary), which is described in a a different document and this doc is probably best to be kept transport agnostic.

oskarth (approved):

Issue #61, opened 2019-11-14
Replace forked Whisper usage with Waku/0 in specs


oskarth:
# Problem
Currently we use a forked version of Whisper that isn't documented in any great detail. This makes it difficult for other clients to implement and get compatibility with Status clients, e.g. status-nim. It also creates confusion around what requirements and guarantees we have, since we are de facto doing more than EIP627 and this has implications around UX, privacy, DDoS resistance, etc (mailservers, light node, rate limiting).

# Acceptance criteria

* [ ] Status Client spec reference Waku spec
* [ ] Updated Waku mode usage spec (replace Whisper Usage and Mailserver)

As well as issues/dependencies marked out for for :

* [x] Current specs for v1 marked clearly as such
* [ ] Deprecation/recommendation notice for Waku mode for client implementers (e.g. status-nim)

# Details
See https://github.com/vacp2p/specs/issues/27 for more details.

This issue is largely dependent on above, but some work can be done in parallel to ease that work. I.e. by documenting customization to Whisper in our current specs, such as:

* [x] mailserver usage
* [x] light mode
* [x] extra packet codes
* [x] rate limiting
* [x] any other changes we have done to EIP627 that aren't accurately captured in our specs (litmus test: 100% compatibility with status-nim w/o reading status-go/whisper code)

# Possible solutions
See acceptance criteria.

# Notes
-

oskarth:
Blocked by https://github.com/vacp2p/specs/issues/27

oskarth:
See https://github.com/vacp2p/pm/issues/7 https://github.com/status-im/specs/issues/73

oskarth:
How is this issue going @samyoul?

cc @andremedeiros

Samyoul:
Hey. I'll have a PR for this by the end of the day.

Samyoul:
@oskarth , the initial rewrite PR draft will not be ready today, there are too many refs to outdated things, or things that are sort of relevant or things that need further research to see if there are Waku analogous versions of Whisper terminology etc.

I will open a draft PR of what I've started, reference it here, and I'll keep adding to it until it is complete ready for review. I naively thought it would be a mostly find and replace job. Most of it is, the minority that will take time, isn't.

Samyoul:
This is the draft https://github.com/status-im/specs/pull/114

Samyoul:
Finally merged https://github.com/status-im/specs/pull/114

PR #114, opened 2020-05-07
Update/waku replace

Samyoul:
Resolving issue https://github.com/status-im/specs/issues/61


docs/stable/1-client.md
```diff
@@ -94,20 +96,24 @@ To communicate between Status nodes, the [RLPx Transport
 Protocol, v5](https://github.com/ethereum/devp2p/blob/master/rlpx.md) is used, which
 allows for TCP-based communication between nodes.
 
-On top of this we run the RLPx-based subprotocol [Whisper
-v6](https://eips.ethereum.org/EIPS/eip-627) for privacy-preserving messaging.
+On top of this we run the RLPx-based subprotocol [Waku
+V1](https://github.com/vacp2p/specs/blob/master/specs/waku/waku-1.md) for privacy-preserving messaging.
 
 There MUST be a node that is capable of discovering peers and
```
oskarth: This isn't accurate, for v1 of the app we still support Whisper. Instead, it should be MAY for Whisper and SHOULD for Waku. I also think the version in production is v1, but I might be wrong on that.

docs/stable/1-client.md
```diff
@@ -94,20 +96,24 @@ To communicate between Status nodes, the [RLPx Transport
...
 
 There are four types of node roles:
 1. Bootstrap nodes
-2. Whisper relayers
+2. Waku relayers
 3. Mailservers (servers and clients)
 4. Mobile nodes (Status Clients)
 
+```js
+// TODO do waku relayers exist? Are they different from mobile nodes?
```
oskarth: Should be essentially same as for Whisper. It'd be good if we kept Whisper in history here, so we have Whisper and Waku relayers. Then, at a later stage, we can deprecate Whisper completely.

cc @jakubgs re Waku relays in cluster

docs/stable/1-client.md
```diff
@@ -165,9 +171,13 @@ It uses Ethereum Node Records (ENR) to report discovered peers.
 Both peers discovery mechanisms use topics to provide peers with certain capabilities.
 There is no point in returning peers that do not support a particular protocol.
 Status nodes that want to be discovered MUST register to Discovery V5 and/or Rendezvous
-with the `whisper` topic. Status nodes that are mail servers and want to
+with the `waku` topic. Status nodes that are mail servers and want to
```
oskarth: See above, MUST seems like too strong of a word and would also make this spec incompatible with v1 in wild, which is not what we want / are doing

docs/stable/1-client.md
```diff
@@ -190,30 +200,30 @@ group of peers the official Status client uses. The others are test networks.
...
 
-### Transport privacy and Whisper usage
+### Transport privacy and Waku usage
 
-Once a Whisper node is up and running there are some specific settings required
+Once a Waku node is up and running there are some specific settings required
 to commmunicate with other Status nodes.
 
-See [3/WHISPER-USAGE](https://specs.status.im/spec/3) for more details.
+See [3/WAKU-USAGE](https://specs.status.im/spec/3) for more details.
```
oskarth: This spec doesnt exist

oskarth (commented):
Good start! I think the main thing to keep in mind is that this spec for Status clients still supports Whisper. That is, we can't just replace Waku with Whisper, since that would indicate clients aren't backwards compatible. Instead, we have to find a way to communicate intent that Whisper MAY be used and Waku SHOULD be used, and that the latter is strongly recommended. Then find a way to refer to that consistently from other parts of the spec.

docs/stable/1-client.md
```diff
@@ -259,9 +269,9 @@ For transaction support, Status clients currently have to rely on Infura.
 
 Status clients currently do not offer native support for file storage.
 
-#### Why do you use Whisper?
+#### Why do you use Waku?
 
-Whisper is one of the [three parts](http://gavwood.com/dappsweb3.html) of the
+Waku is one of the [three parts](http://gavwood.com/dappsweb3.html) of the
```
kdeme: I guess this should remain Whisper.

docs/stable/2-account.md
```diff
@@ -81,19 +80,23 @@ Everything else associated with the contact is either verified or derived from t
     - Timestamp
 - These bundles are made available in a variety of ways, as defined in section 2.1.
 
+```js
+// TODO is there an "Open Waku Systems" ^^^?
```
kdeme: Open Whisper Systems has nothing to do with the Whisper protocol.

docs/stable/3-waku-usage.md
```diff
@@ -1,89 +1,98 @@
...
-| P2P Request | 126 | ✔ | [4/WHISPER-MAILSERVER](https://specs.status.im/spec/4) |
-| P2P Messages | 127 | ✔/𝘅 (EIP-627 supports only single envelope in a packet) | [4/WHISPER-MAILSERVER](https://specs.status.im/spec/4) |
-
+## Waku packets
+
+| Packet Name          | Code | References |
+| -------------------- | ---: | --- |
+| Status               |    0 | [Handshake](#handshake), [WAKU-1](https://github.com/vacp2p/specs/blob/master/specs/waku/waku-1.md#status) |
+| Messages             |    1 | [WAKU-1](https://github.com/vacp2p/specs/blob/master/specs/waku/waku-1.md#messages) |
+| PoW Requirement      |    2 | [WAKU-1](https://github.com/vacp2p/specs/blob/master/specs/waku/waku-1.md#pow-requirement-update) |
```
kdeme: 2,3, 123 & 124 are not part of Waku spec, and thus not used afaik.
See updates here: https://github.com/vacp2p/specs/pull/119 (and if they get accepted / corrected)

docs/stable/1-client.md
Samyoul: Ok. Thank you.

docs/stable/1-client.md
Samyoul: My presumption is that this will reference https://github.com/status-im/specs/blob/master/docs/stable/3-whisper-usage.md, after the whole rewrite has completed.

docs/stable/1-client.md
Samyoul: Yeah. I wasn't sure if we were treating Waku as the direct successor and replacement for Whisper, but it doesn't make sense in this sentence. I think it would be best to revert the change and remove this sentence when we phase Whisper our 100%.

docs/stable/2-account.md
Samyoul: Ah, thank you. I'll add a reference to https://en.wikipedia.org/wiki/Signal_Messenger#2013%E2%80%932018:_Open_Whisper_Systems

docs/stable/3-waku-usage.md
Samyoul: Yeah. This section needs a rewrite after our chat yesterday https://discordapp.com/channels/624205794384281629/626744455361331222/707997524631355464

Samyoul:
@oskarth @kdeme Should these specs include a page or at least a section about the Whisper / Waku bridge implementation?

Currently we have the functionality but it isn't documented. Whisper / Waku bridging is [mentioned in the specs](https://github.com/vacp2p/specs/blob/master/specs/waku/waku-1.md#waku-whisper-bridging), but there is no mention (at least that I've seen) of bridging being required. I believe that we need clear language that details that a Status network simultaneously running Whisper and Waku MUST have a Whisper / Waku bridge, to allow messages to propagate both networks. Otherwise supporting both Whisper and Waku becomes redundant.

Samyoul:
To properly update this spec we will need at least 2 additional pages.

- Waku usage version of https://github.com/status-im/specs/blob/master/docs/stable/3-whisper-usage.md
- Waku mailserver version of https://github.com/status-im/specs/blob/master/docs/stable/4-whisper-mailserver.md

This will allow for detailing the Whisper usage concurrently with detailing the Waku usage.

Would the below file names match the current convention?

- 1-client.md
- 2-account.md
- 3-whisper-usage.md
- 4-whisper-mailserver.md
- 5-secure-transport.md
- 6-payloads.md
- 8-eips.md
- **9-waku-usage.md <**
- **10-waku-mailserver.md <**

docs/stable/1-client.md
Samyoul: I'm going to revert this change, as per https://github.com/status-im/specs/pull/114#issuecomment-625894903

docs/stable/1-client.md
```diff
@@ -94,33 +96,39 @@ To communicate between Status nodes, the [RLPx Transport
...
 #### Node discovery and roles
 
-There are four types of node roles:
-1. Bootstrap nodes
-2. Whisper relayers
-3. Mailservers (servers and clients)
-4. Mobile nodes (Status Clients)
+There are five types of node roles:
+1. `Bootstrap node`
+1. `Waku relayer`
```
kdeme: Probably shouldn't create a new role for Waku relayer. Just call it Waku or Whisper Relayer (or just Relayer?).
As In theory then you also have to split Mailserver and Mobile node, they both still us the Waku / Whisper protocol

docs/stable/1-client.md
```diff
@@ -94,33 +96,39 @@ To communicate between Status nodes, the [RLPx Transport
...
+1. `Whisper relayer`
+1. `Mailserver` (servers and clients)
+1. `Mobile node` (Status Clients)
 
-To implement a standard Status client you MUST implement both 2. and 4. node types. The
-other node types are optional, but we RECOMMEND you implement a mailserver
-client mode, otherwise the user experience is likely to be poor.
+To implement a standard Status client you MUST implement `Waku relayer` and/or `Whisper relayer` roles,
+additionally the Status client MUST implement the `Mobile node` role.
```
kdeme: I'm questioning the original sentence here. I think `Waku relayer` node is enough to implement a Status client. From the small explanation that is written about mobile nodes, I conclude this is the Waku light node. So, why would it not work with the full node, the light node is a subset.

docs/stable/1-client.md
```diff
@@ -338,6 +366,10 @@ A mailserver has to be online to receive messages for other nodes, this puts a h
 
 Use of gossip-based routing doesn't necessarily scale. It means each node can see a message multiple times, and having too many light nodes can cause propagation probability that is too low. See [Whisper vs PSS](https://our.status.im/whisper-pss-comparison/) for more and a possible Kademlia based alternative.
 
+```js
+// TODO investigate update or Waku analogous ^^^
```
kdeme: Nothing has changed in this aspect for Waku

docs/stable/1-client.md
```diff
@@ -94,33 +96,39 @@ To communicate between Status nodes, the [RLPx Transport
...
-v6](https://eips.ethereum.org/EIPS/eip-627) for privacy-preserving messaging.
+On top of this we run RLPx-based subprotocols, the client
+MAY use [Whisper v6](https://eips.ethereum.org/EIPS/eip-627), though the client 
+SHOULD use [Waku V1](https://github.com/vacp2p/specs/blob/master/specs/waku/waku-1.md)
+for privacy-preserving messaging.
 
-There MUST be a node that is capable of discovering peers and
-implements Whisper V6 specification.
+There MUST be a node that is capable of discovering peers and implements
+[Whisper v6](https://eips.ethereum.org/EIPS/eip-627),
```
kdeme: The way it is written now sounds like you MUST implement all 3. While it is actually only 1.

docs/stable/1-client.md
Samyoul: In this section I was implementing a suggestion from @oskarth https://github.com/status-im/specs/pull/114#discussion_r421938443 .

While rewriting this section I had the idea to simply rename the `Whisper relayer` to `Message relayer` and add a definition of a`Message relayer`. I decided to explicitly keep both `Whisper relayer` and `Waku relayer` so that deprecation would be simpler.

Yes I may eventually split the `Mailserver`, but I haven't rewritten that section yet and I wanted to see how divergent Whisper vs Waku Mailservers were.

Could you explain why the `Mobile node` would need to be split?

docs/stable/1-client.md
Samyoul: I intentionally wrote this so that there "MUST be **a** node" that implements all 3. I presumed that if there is a network that MAY implement Whisper, Waku 0 and Waku 1, then that network must have at least 1 node that implements all messaging protocols.

The other alternative is to make the language conditional and state that "There MUST be a node that ... implements all the messaging protocols that the network supports".

This would probably require a section defining the term "messaging protocol".

docs/stable/1-client.md
Samyoul: Should we have a concept of a `Mobile node`? Currently a `mobile node` is defined as:

>This is a Whisper and/or Waku node which connects to part of the respective Whisper
and/or Waku network(s). It MAY relay messages.

How does this diverge from a plain Whisper / Waku node ?

docs/stable/1-client.md
kdeme: I wouldn't split them, as the role is essentially the same, just a different protocol. As is the case for Mail server and mobile node. Yeah, perhaps Message relayer (Waku or Whisper) is better. Or just Waku/Whisper relayer.

docs/stable/1-client.md
kdeme: From the current description, or rather from the fact that "Status clients" is mentioned, it looks like the mobile node role is a node that not only speaks Waku/Whisper, but also the rest of the Status protocol to be able to send/receive chat messages (the layers on top of Whisper/Waku).

Compared to a Waku/Whisper relayer which doesn't need to know these layers to just pass the Waku/Whisper envelopes. I don't fully like the name mobile node, but ok, that is not the scope of this PR. So yeah, you can ignore my comment.

docs/stable/1-client.md
kdeme: Right, I understand now, but this was not clear to me when I first read it.
Yes, I think it would be better to explain that part in another section where e.g. the bridging is also explained, or at least referenced from the vacp2p Waku spec.
Perhaps this part does not even belong in the client.md, as you do not have to do this to create a Status client.

docs/stable/1-client.md
Samyoul: I like `Waku/Whipser relayer`, let's go with that.

docs/stable/1-client.md
Samyoul: >Perhaps this part does not even belong in the client.md, as you do not have to do this to create a Status client.

Yes, I agree with this. Although `client.md` is home to a lot of details about different types of nodes required for a Status client to operate effectively.

[I think it is necessary for a Whisper / Waku bridge spec page](https://github.com/status-im/specs/pull/114#issuecomment-625887093), although the bridge spec would be fairly small, the details of the bridge spec would not neatly belong in any of the currently existing or proposed spec pages.

It would also allow for easy deprecation when required. I don't know if that would be overkill. What do you think?

docs/stable/1-client.md
```diff
@@ -94,33 +96,39 @@ To communicate between Status nodes, the [RLPx Transport
...
+1. `Bootstrap node`
+1. `Waku relayer`
+1. `Whisper relayer`
+1. `Mailserver` (servers and clients)
+1. `Mobile node` (Status Clients)
 
-To implement a standard Status client you MUST implement both 2. and 4. node types. The
-other node types are optional, but we RECOMMEND you implement a mailserver
-client mode, otherwise the user experience is likely to be poor.
+To implement a standard Status client you MUST implement `Waku relayer` and/or `Whisper relayer` roles,
```
decanus:
```suggestion
A standard Status client MUST implement `Waku relayer` and/or `Whisper relayer` roles,
```

I don't like the usage of you

docs/stable/1-client.md
```diff
@@ -94,33 +96,39 @@ To communicate between Status nodes, the [RLPx Transport
...
+additionally the Status client MUST implement the `Mobile node` role.
+The other node types are optional, but we RECOMMEND you implement a `Mailserver`
```
decanus:
```suggestion
The other node types are optional, but it is RECOMMEND to implement a `Mailserver`
```

decanus (changes requested):

docs/stable/1-client.md
Samyoul: Neither do I.  There's quite a bit of the use of personal pronouns in these specs. I'll keep an eye out for them and replace them as I come across them.

docs/stable/1-client.md
Samyoul: Probably should say 

The other node types are optional, but it is RECOMMENDED to implement a `Mailserver`

Samyoul:
@decanus removing all personal pronouns from these specs will change the whole tone of the writing.  Is this something we want to introduce with this draft?

docs/stable/1-client.md
Samyoul: @kdeme what is your opinion?

Samyoul:
Do we want to update the [8-eips.md](https://github.com/status-im/specs/blob/master/docs/stable/8-eips.md) with the Vac p2p specs for Waku? 

Would a document dedicated to BIPs and EIPs be the correct place for the Vac specs?

If yes, should we rename this document from `8-eips.md` to `8-supported-specs.md`?

oskarth:
@Samyoul please mark this PR as 'ready to review' once it is ready to be reviewed.

> Do we want to update the 8-eips.md with the Vac p2p specs for Waku?

> Would a document dedicated to BIPs and EIPs be the correct place for the Vac specs?

> If yes, should we rename this document from 8-eips.md to 8-supported-specs.md?

This seems unrelated to the issue this PR should solve.

Samyoul:
@oskarth no problem I will, but I've still got a few items to address and then I'll have some questions that need answering.

---

> This seems unrelated to the issue this PR should solve.

The issues says:

> Status Client spec reference Waku spec

I've included references to the Waku specs in every document that requires it, with the exception of `8-eips.md`. From my understanding `8-eips.md` should also have references to Waku specs, and if `8-eips.md` has references to Waku specs the document name should change (In fact `8-eips.md` already should change its name as it also contains BIPs).

To me this change is entirely in the scope of the issue's requirements. If you would be happier with this particular change in a separate PR I will open one later today.

docs/stable/1-client.md
kdeme: Either works for me. Could also be just a subsection in the Whisper-usage.md. As it is bridging to support still Whisper. And can be removed easily then later on.

decanus:
> @decanus removing all personal pronouns from these specs will change the whole tone of the writing. Is this something we want to introduce with this draft?

I'd be fine with doing this later on.

oskarth:
The intent of https://github.com/status-im/specs/blob/master/docs/stable/8-eips.md was more to indicate various EIPs/(BIPs) not captured elsewhere in detail. Waku isn't an EIP, but I suppose we could add it there as opposed to only having it in-line. Up to you. In that case, it'd probably be more accurate to extend title of EIP to something more general (cc @3esmit). Just beware of scope creep :)

3esmit:
Notice that Whisper is a EIP, see https://eips.ethereum.org/EIPS/eip-627 and Waku seems to fall into the same preposition.

Samyoul:
I've split this issue into https://github.com/status-im/specs/issues/116

Samyoul:
### Questions

@oskarth, @decanus, @kdeme, @cammellos 

---

https://github.com/status-im/specs/blob/87cab77f203efa7d53a19ce47bf7624255804eb4/docs/stable/1-client.md#L168-L170

- Is there a `Waku` analogous term of `whispermail`?

---

https://github.com/status-im/specs/blob/2c7d086e0973667a929cc3ce5a18408ab5959bfe/docs/stable/9-waku-usage.md#L292-L297

- Is there a Waku analogous version of `shh_generateSymKeyFromPassword`?

---

https://github.com/status-im/specs/blob/2c7d086e0973667a929cc3ce5a18408ab5959bfe/docs/stable/9-waku-usage.md#L328-L330

- Does Waku use extensions. Are the extensions now part of the Waku spec?

kdeme:
@Samyoul I'll reply with what I know. But others might want to chime in.

>     * Is there a `Waku` analogous term of `whispermail`?
I don't know this. I'm not even sure if we still use the (old) discovery v5 & its topics registration in combination with Waku.

>     * Should this reference to [the theoretical scaling model](https://github.com/vacp2p/research/tree/dcc71f4779be832d3b5ece9c4e11f1f7ec24aac2/whisper_scalability) be retained with the introduction of Waku?
That part of the scaling problem has been tackled with `topic-interest` in Waku. However, there are other scalability issues (mailserver and the general gossip/flooding nature of the protocol)

>     * If so should this section be rewritten from a post-Whisper perspective?
Perhaps yes, based on the above.
> 
>     * Has research been done and/or introduced into Waku?
Not really no.
> 

>     * ~Should I add a table for the message Status Options? See the items updated here~ [~Clarify packet types and fix status-options fields sections~](https://github.com/vacp2p/specs/pull/119)
Yeah, I'm not sure why we need to repeat that part to begin with.

>     * Is there a Waku analogous version of `shh_generateSymKeyFromPassword`?
I'd guess there is such JSON-RPC call. But be aware that RPC calls are currently not specified at all in Waku. So if it is there it is probably copied from the Whisper v6 JSON-RPC calls.

>     * Does Waku use extensions. Are the extensions now part of the Waku spec?
Not part of the spec. Not sure what is used.

Samyoul:
I've opened this PR for review. Please note that I've still got 3 outstanding points that I feel need clarification https://github.com/status-im/specs/pull/114#issuecomment-629272782

docs/stable/9-waku-usage.md
```diff
@@ -0,0 +1,346 @@
...
+| P2P Request          |  126 | [4/WAKU-MAILSERVER](https://specs.status.im/spec/4), [WAKU-1](https://github.com/vacp2p/specs/blob/master/specs/waku/waku-1.md#p2p-request) |
+| P2P Messages         |  127 | [4/WAKU-MAILSERVER](https://specs.status.im/spec/4), [WAKU-1](https://github.com/vacp2p/specs/blob/master/specs/waku/waku-1.md#p2p-message) |
+
+## Waku node configuration
+
+If you want to run a Waku node and receive messages from Status clients, it must be properly configured.
+
+Waku's Proof Of Work algorithm is used to deter denial of service and various spam/flood attacks against the Waku network. The sender of a message must perform some work which in this case means processing time. Because Status' main client is a mobile client, this easily leads to battery draining and poor performance of the app itself. Hence, all clients MUST use the following Waku node settings:
+* proof-of-work requirement not larger than `0.002` for payloads less than 50,000 bytes
+* proof-of-work requirement not larger than `0.000002` for payloads greater than or equal to 50,000 bytes
```
Samyoul, line 82: This change may be superseded by https://github.com/status-im/specs/pull/118/files . @cammellos what do you think? 

For clarification I've included this in the new `9-waku-usage.md` doc, and I can see your change is to the `3-whisper-usage.md`. I have not made any material changes myself to `3-whisper-usage.md`

docs/stable/1-client.md
```diff
@@ -344,14 +344,6 @@ There are several security considerations to take into account when running Stat
 
 ### Scalability and UX
 
-**Bandwidth usage:**
-
-In version 1 of Status, bandwidth usage is likely to be an issue. For more investigation into this, see [the theoretical scaling model](https://github.com/vacp2p/research/tree/dcc71f4779be832d3b5ece9c4e11f1f7ec24aac2/whisper_scalability).
```
oskarth, line 335: I'd keep this as it is still relevant in version 1. Replace second sentence with "In version <X> this is partially addressed with Waku usage

docs/draft/7-group-chat.md
```diff
@@ -20,13 +20,13 @@ title: 7/GROUP-CHAT
...
   - [Signature](#signature)
   - [Group membership event](#group-membership-event)
-    - [chat-created](#chat-created)
-    - [name-changed](#name-changed)
-    - [members-added](#members-added)
-    - [members-joined](#members-joined)
-    - [admins-added](#admins-added)
-    - [members-removed](#members-removed)
-    - [admin-removed](#admin-removed)
+    - [CHAT_CREATED](#chat_created)
```
oskarth: Changes to this file seems unrelated to this PR.

docs/stable/1-client.md
```diff
@@ -94,33 +96,33 @@ To communicate between Status nodes, the [RLPx Transport
...
 #### Node discovery and roles
 
-There are four types of node roles:
-1. Bootstrap nodes
-2. Whisper relayers
-3. Mailservers (servers and clients)
-4. Mobile nodes (Status Clients)
+There are five types of node roles:
+1. `Bootstrap node`
+1. `Whisper/Waku relayer`
```
oskarth, line 109: Numbered list format off

oskarth: Also where is the five from?

docs/stable/1-client.md
```diff
@@ -94,33 +96,33 @@ To communicate between Status nodes, the [RLPx Transport
...
+1. `Mailserver` (servers and clients)
+1. `Mobile node` (Status Clients)
 
-To implement a standard Status client you MUST implement both 2. and 4. node types. The
-other node types are optional, but we RECOMMEND you implement a mailserver
-client mode, otherwise the user experience is likely to be poor.
+A standard Status client MUST implement `Whisper/Waku relayer` roles,
```
oskarth: Hmm I realize this was in old code, but is this true? If you implement a mobile node, do you also need to implement a relay node? It is true we get forwarded envelopes on mobile (but don't relay further). What do we mean by mobile node in that case? Something seems off here

cc @kdeme @cammellos @decanus

docs/stable/1-client.md
```diff
@@ -165,15 +167,19 @@ It uses Ethereum Node Records (ENR) to report discovered peers.
 Both peers discovery mechanisms use topics to provide peers with certain capabilities.
 There is no point in returning peers that do not support a particular protocol.
 Status nodes that want to be discovered MUST register to Discovery V5 and/or Rendezvous
-with the `whisper` topic. Status nodes that are mail servers and want to
+with the `whisper` and/or `waku` topic(s). Status nodes that are `Mailservers` and want to
```
oskarth: Did we definitely change this to `waku` topic?

docs/stable/1-client.md
```diff
@@ -94,33 +96,33 @@ To communicate between Status nodes, the [RLPx Transport
...
 
-On top of this we run the RLPx-based subprotocol [Whisper
-v6](https://eips.ethereum.org/EIPS/eip-627) for privacy-preserving messaging.
-
-There MUST be a node that is capable of discovering peers and
-implements Whisper V6 specification.
+On top of this we run RLPx-based subprotocols, the client
+MAY use [Whisper v6](https://eips.ethereum.org/EIPS/eip-627), though the client 
+SHOULD use [Waku V1](https://github.com/vacp2p/specs/blob/master/specs/waku/waku-1.md)
+for privacy-preserving messaging.
```
oskarth: Perhaps add a sentence on why Waku should be used, like "Waku should be used because it preserves your bandwidth better". Additionally, we might want to say here that Whisper SHOULD NOT be used to indicate that it is a compatibility thing.

docs/stable/1-client.md
```diff
@@ -190,30 +196,33 @@ group of peers the official Status client uses. The others are test networks.
...
-secrecy of messages we implement a secure transport on top of Whisper. This is
+secrecy of messages we implement a secure transport on top of Whisper and Waku. This is
 used in 1:1 chats and group chats, but not for public chats. See [5/SECURE-TRANSPORT](https://specs.status.im/spec/5) for more.
 
 ### Data Sync
 
 [MVDS](https://specs.vac.dev/mvds.html) is used for 1:1 and group chats, however it is currently not in use for public chats.
 
-[Status payloads](#payloads-and-clients) are serialized and then wrapped inside a MVDS message which is added to an [MVDS payload](https://specs.vac.dev/mvds.html#payloads), this payload is then encrypted (if necessary for 1-to-1 / group-chats) and sent using whisper which also encrypts it.
+[Status payloads](#payloads-and-clients) are serialized and then wrapped inside a
```
oskarth: spacing

docs/stable/1-client.md
```diff
@@ -232,7 +241,7 @@ See [Appendix A](#appendix-a-security-considerations)
 
 ## Design Rationale
 
-### P2P Overlay
+### P2P Overlay 1
```
oskarth: Extra number?

docs/stable/1-client.md
```diff
@@ -297,11 +316,11 @@ Discovery v5 for a short period of time until their peer list is populated.
...
 [here](https://github.com/status-im/swarms/blob/master/ideas/092-disc-v5-research.md).
 
-#### I heard something about mailservers being trusted somehow?
+#### I heard something about `Mailservers` being trusted somehow?
 
-In order to use a mail server, a given node needs to connect to it directly, i.e. add the mailserver as its peer and mark it as trusted. This means that the mail server is able to send direct p2p messages to the node instead of broadcasting them. Effectively, it knows the bloom filter of the topics the node is interested in, when it is online as well as many metadata like IP address.
+In order to use a `Mailserver`, a given node needs to connect to it directly, i.e. add the `Mailserver` as its peer and mark it as trusted. This means that the `Mailserver` is able to send direct p2p messages to the node instead of broadcasting them. Effectively, it knows the bloom filter of the topics the node is interested in, when it is online as well as many metadata like IP address.
 
-### Data sync
+### Data sync 1
```
oskarth: Extra number?

docs/stable/1-client.md
```diff
@@ -352,13 +366,13 @@ The main privacy concern with light nodes is that directly connected peers will
 
 By having a bloom filter where only the topics you are interested in are set, you reveal which messages you are interested in. This is a fundamental tradeoff between bandwidth usage and privacy, though the tradeoff space is likely suboptimal in terms of the [Anonymity](https://eprint.iacr.org/2017/954.pdf) [trilemma](https://petsymposium.org/2019/files/hotpets/slides/coordination-helps-anonymity-slides.pdf).
 
-**Mailserver client privacy:**
+**`Mailserver` client privacy:**
 
-A mailserver client has to trust a mailserver, which means they can send direct traffic. This reveals what topics / bloom filter a node is interested in, along with its peerID (with IP).
+A `Mailserver` client has to trust a `Mailserver`, which means they can send direct traffic. This reveals what topics / bloom filter a node is interested in, along with its peerID (with IP).
```
oskarth: Is there a reason Mailserver is in `` but `mailserver client` isn't?

docs/stable/9-waku-usage.md
cammellos: It's probably best to keep them separate, yes, if you merge in this first I will rebase against it, and update `9-waku-usage.md` as well

docs/stable/1-client.md
```diff
@@ -311,8 +330,7 @@ whereby participants can sync. Additionally, MVDS is currently not optimized for
 large group contexts, which means bandwidth usage will be a lot higher than
 reasonable. See [P2P Data Sync for
 Mobile](https://vac.dev/p2p-data-sync-for-mobile) for more. This is an active
-area of research. The bandwidth issue is further exacerbated by Whisper being
-very bandwidth heavy.
+area of research.
```
oskarth, line 332: Could we add a CHANGELOG and bump version number somewhere here too?

docs/stable/3-whisper-usage.md
```diff
@@ -24,14 +24,16 @@ title: 3/WHISPER-USAGE
     - [Contact code topic](#contact-code-topic)
     - [Partitioned topic](#partitioned-topic)
     - [Public chats](#public-chats)
-    - [Generic discovery topic](#generic-discovery-topic)
-    - [One-to-one topic](#one-to-one-topic)
+    <!-- - [Generic discovery topic](#generic-discovery-topic) -->
```
oskarth: Why are these commented out?

docs/stable/5-secure-transport.md
```diff
@@ -47,12 +47,8 @@ It builds on the [X3DH](https://signal.org/docs/specifications/x3dh/) and [Doubl
     - [Multi-device support](#multi-device-support)
     - [Pairing](#pairing)
     - [Sending messages to a paired group](#sending-messages-to-a-paired-group)
-    - [Account recovery](#account-recovery)
+    - [Account recovery](#account-recovery-1)
```
oskarth: Unrelated change / extra number?

docs/stable/6-payloads.md
```diff
@@ -14,39 +14,39 @@ title: 6/PAYLOADS
 
 ## Abstract
 
-This specifications decribes how the payload of each message in Status looks
-like. It is primarly centered around chat and chat-related use cases.
+This specification describes how the payload of each message in Status looks
```
oskarth: Unrelated change

docs/stable/8-eips.md
```diff
@@ -17,7 +17,7 @@ title: 8/EIPS
 In this specification, we describe how Status relates with EIPs.
 
 ## Table of Contents
-- [Status EIPs standards ](#status-client-specification)
+- [Status EIPs standards ](#8eips)
```
oskarth: Unrelated change?

oskarth (changes requested):
Partial review, might have more comments later.

Generally lgtm - a lot more changes than I thought it'd be :sweat_smile: 

For the unrelated changes, I'd put that in a separate PR as general typo/toc fixes. That way it is much easier to read git history/revert etc.

docs/stable/1-client.md
Samyoul: Ah good catch, I had previously split Whisper and Waku relayers into separate node roles. I will change this.

I changed the numbered list so that all are `1.` because markdown parses `1.` as an ordered list. That means that additions to the list don't need to be manually re-sequenced.

docs/stable/1-client.md
Samyoul: That makes more sense.

docs/stable/1-client.md
Samyoul: I don't know. It is related to one of my outstanding issues. There is a reference in the specs to `whispermail`, but I haven't been able to find if it was changed to `wakumail` with Waku. 

In this section I presumed that Waku would have its `waku` topic set rather than still used the `whisper` topic.

docs/stable/1-client.md
Samyoul: These things are my IDE. It forces unique anchors. I'll change it.

docs/stable/3-whisper-usage.md
Samyoul: The sections they refer to are also commented out. Looks like whoever commented out the sections left in the links to the sections. I presumed that if the sections had been commented out that they were being retained for some purpose, and so instead of removing the links I commented them out also.

docs/stable/1-client.md
Samyoul: FYI, I also added a reference to this in the new `9-waku-usage.md` document, under the `topic-interest` references.

docs/stable/1-client.md
kdeme: Yeah, I guess it should rather be something like:
"A Status client MUST implement the `Mobile node` role and MAY implement `Whisper/Waku relayer` and `Mailserver` roles."

kdeme: The whole role thing is not so clear though. I think it is because we are mixing Whisper/Waku & Status Client layers. And then also bootstrap..

docs/stable/1-client.md
Samyoul: ~~A point of clarification. Are you suggesting to change:~~

~~"MAY use Whisper" to "SHOULD NOT use Whisper"?~~

EDIT:

I'm presuming that is what you mean.

docs/stable/1-client.md
Samyoul: >If you implement a mobile node, do you also need to implement a relay node?

I puzzled over this question when I was drafting this PR.

>"A Status client MUST implement the Mobile node role and MAY implement Whisper/Waku relayer and Mailserver roles."

Are there circumstances when a Status node would be active and not be using p2p messaging?

docs/stable/5-secure-transport.md
Samyoul: This is the way GitHub differentiates between anchors of the same name.

The first instance is `#anchor-name` the nth additional instance is referenced `#anchor-name-n` where `n` is the order in which subsequent anchors are placed in the md document.

I'll move these out to #120

docs/stable/3-whisper-usage.md
Samyoul: I'm moving these into #120

docs/stable/1-client.md
oskarth: Huh TIL. So this will render correctly for all (well, ours) kramdown generator?

docs/stable/1-client.md
oskarth: Don't know. I'd grep the status-go code for this. If you can only find whispermail we didn't change it.

docs/stable/1-client.md
oskarth: Yeah, we need to clarify the roles a lot more. Since this issue exists in previous spec, could you open an issue for clarifying this @Samyoul? Then we can deal with it separately to get this merged. (I think there was some related roles issue but can't find it right now)

docs/stable/1-client.md
```diff
@@ -190,30 +194,33 @@ group of peers the official Status client uses. The others are test networks.
...
-secrecy of messages we implement a secure transport on top of Whisper. This is
+secrecy of messages we implement a secure transport on top of Whisper and Waku. This is
 used in 1:1 chats and group chats, but not for public chats. See [5/SECURE-TRANSPORT](https://specs.status.im/spec/5) for more.
 
 ### Data Sync
 
 [MVDS](https://specs.vac.dev/mvds.html) is used for 1:1 and group chats, however it is currently not in use for public chats.
 
-[Status payloads](#payloads-and-clients) are serialized and then wrapped inside a MVDS message which is added to an [MVDS payload](https://specs.vac.dev/mvds.html#payloads), this payload is then encrypted (if necessary for 1-to-1 / group-chats) and sent using whisper which also encrypts it.
+[Status payloads](#payloads-and-clients) are serialized and then wrapped inside an
```
oskarth, line 220: This looks weird spacing-wise in GH at least

docs/stable/6-payloads.md
```diff
@@ -200,9 +200,9 @@ This will satisfy the Lamport requirement, namely: a -> b then T(a) < T(b)
 
 `clock` SHOULD be calculated using the algorithm of [Lamport timestamps](https://en.wikipedia.org/wiki/Lamport_timestamps). When there are messages available in a chat, `clock`'s value is calculated based on the last received message in a particular chat: `max(timeNowInMs, last-message-clock-value + 1)`. If there are no messages, `clock` is initialized with `timestamp`'s value.
 
-Messages with a `clock` greater than `120` seconds over the whisper timestamp SHOULD be discarded, in order to avoid malicious users to increase the `clock` of a chat arbitrarily.
+Messages with a `clock` greater than `120` seconds over the whisper/Waku timestamp SHOULD be discarded, in order to avoid malicious users to increase the `clock` of a chat arbitrarily.
```
oskarth: Consistent capitalization

docs/stable/9-waku-usage.md
```diff
@@ -0,0 +1,346 @@
...
+title: 9/WAKU-USAGE
+---
+
+# 9/WAKU-USAGE
+
+> Version: 0.2
+>
+> Status: Stable
+>
+> Authors: Adam Babik <adam@status.im>, Corey Petty <corey@status.im>, Oskar Thorén <oskar@status.im> (alphabetical order)
```
oskarth: Add your name here

oskarth:
If we resolve outstanding issues and fix conflicting files it lgtm

docs/stable/1-client.md
oskarth: https://github.com/status-im/status-go/search?q=whispermail&unscoped_q=whispermail and https://github.com/status-im/status-go/search?q=wakumail&unscoped_q=wakumail seems like no

docs/stable/1-client.md
Samyoul: I had to check this, but it appears to already be something some contributors have done, and is working with the live specs.

See https://specs.status.im/spec/2#introduction

> 1 a whisper chat identity keypair
> 2 a set of cryptocurrency wallet keypairs

Which in the raw `.md` file looks like this :

https://github.com/status-im/specs/blob/56d5a2ff3949f3d2a68c70e1ae6780d2a080f16d/docs/stable/2-account.md#L51-L54

---

I just saw a better example:

https://specs.status.im/spec/2#private-11-messages

> 1 scanning a user generated QR code
> 2 discovery through the Status app
> 3 asynchronous X3DH key exchange
> 4 public key via public channel listening
>   . status-react/src/status_im/contact_code/core.cljs
> 5 contact codes
> 6 decentralized storage (not implemented)
> 7 whisper

In raw `.md` file looks like:

https://github.com/status-im/specs/blob/56d5a2ff3949f3d2a68c70e1ae6780d2a080f16d/docs/stable/2-account.md#L130-L139

This is a good example of how Markdown actually parses a list of `1.` mixed with other numbers. It is easier to just have all items as `1.` so no manual sequencing is needed.

decanus:
@Samyoul this PR seems to have gotten rather large, I'd be happy if we merge it soon.

docs/stable/1-client.md
Samyoul: https://github.com/status-im/specs/issues/123

docs/stable/1-client.md
Samyoul: That is a useful feature! Ok, I suppose that answers the "do we" question, the next question would be "should we?".

I'll remove the changes I've made here and I'll also break this into a separate issue.

Samyoul:
@oskarth @decanus I believe that this monster is now ready for the open seas 🐳

docs/stable/2-account.md
```diff
@@ -219,3 +221,13 @@ All messages sent are encrypted with the public key of the destination and signe
 ## Security Considerations
 
 -
+
+## Changelog
+
+### Version 0.3
+
+Released `TODO`
```
oskarth, line 227: Don't forget this one

docs/stable/3-whisper-usage.md
```diff
@@ -329,17 +345,26 @@ It's recommended to return messages from newest to oldest. To move further back
...
-The above `topics` is then converted into a bloom filter and then and sent to the Mailserver.
+The above `topics` is then converted into a bloom filter and then and sent to the `Mailserver`.
 
 <!-- TODO: Clarify actual request with bloom filter to mailserver -->
+
+## Changelog
+
+### Version 0.3
+
+Released `TODO`
```
oskarth, line 364: Same

oskarth (approved):
Nice!

docs/stable/1-client.md
```diff
@@ -325,17 +337,17 @@ very bandwidth heavy.
...
 
-There are several security considerations to take into account when running Status. Chief among them are: scalability, DDoS-resistance and privacy. These also vary depending on what capabilities are used, such as mailserver, light node, and so on.
+There are several security considerations to take into account when running Status. Chief among them are: scalability, DDoS-resistance and privacy. These also vary depending on what capabilities are used, such as `Mailserver`, light node, and so on.
 
 ### Scalability and UX
 
 **Bandwidth usage:**
 
-In version 1 of Status, bandwidth usage is likely to be an issue. For more investigation into this, see [the theoretical scaling model](https://github.com/vacp2p/research/tree/dcc71f4779be832d3b5ece9c4e11f1f7ec24aac2/whisper_scalability).
+In version 1 of Status, bandwidth usage is likely to be an issue. In version this is partially addressed with Waku usage, see [the theoretical scaling model](https://github.com/vacp2p/research/tree/dcc71f4779be832d3b5ece9c4e11f1f7ec24aac2/whisper_scalability).
```
kdeme: I think you for got to add the actual version?

docs/stable/1-client.md
```diff
@@ -95,22 +97,21 @@ To communicate between Status nodes, the [RLPx Transport
...
 allows for TCP-based communication between nodes.
 
-On top of this we run the RLPx-based subprotocol [Whisper
-v6](https://eips.ethereum.org/EIPS/eip-627) for privacy-preserving messaging.
-
-There MUST be a node that is capable of discovering peers and
-implements Whisper V6 specification.
+On top of this we run RLPx-based subprotocols, the client
+SHOULD NOT use [Whisper v6](https://eips.ethereum.org/EIPS/eip-627), the client 
+SHOULD use [Waku V1](https://github.com/vacp2p/specs/blob/master/specs/waku/waku-1.md)
```
kdeme, line 102: nitpick: no capital v like in Whisper v6

kdeme (approved):

docs/stable/1-client.md
Samyoul: Good catch.

docs/stable/10-waku-mailserver.md
```diff
@@ -0,0 +1,129 @@
...
+
+For a requester, to know that all messages have been sent by a `Mailserver`, it SHOULD handle P2P Request Complete code (`0x7d`). This code is followed by the following parameters:
+
+```
+[ RequestID, LastEnvelopeHash, Cursor ]
+```
+
+`RequestID`: 32-byte wide array with a Keccak-256 hash of the envelope containing the original request  
+`LastEnvelopeHash`: 32-byte wide array with a Keccak-256 hash of the last sent envelope for the request  
+`Cursor`: an array of a cursor returned from the previous request (optional)
```
decanus:
```suggestion
 - `RequestID`: 32-byte wide array with a Keccak-256 hash of the envelope containing the original request  
 - `LastEnvelopeHash`: 32-byte wide array with a Keccak-256 hash of the last sent envelope for the request  
 - `Cursor`: an array of a cursor returned from the previous request (optional)
```

docs/stable/2-account.md
Samyoul: I've got it on my Todo list. Once this is merged I'll reference the commit hash in an update PR.

docs/stable/10-waku-mailserver.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Confidentiality
+
+All Waku envelopes are encrypted. A `Mailserver` node can not inspect their contents.
+
+### Altruistic and centralized operator risk
+
+In order to be useful, a `Mailserver` SHOULD be online most of time. That means
+you either have to be a bit tech-savvy to run your own node, or rely on someone
+else to run it for you.
```
decanus:
```suggestion
In order to be useful, a `Mailserver` SHOULD be online most of time. That means
users either have to be a bit tech-savvy to run their own node, or rely on someone
else to run it for them.
```

decanus (changes requested):

docs/stable/10-waku-mailserver.md
Samyoul: I thought we were addressing this type of thing in https://github.com/status-im/specs/issues/115

docs/stable/10-waku-mailserver.md
Samyoul: Should the same change also be applied to `4-whisper-mailserver.md`?

decanus: probably, but not in this PR probably.

Samyoul:
Hey @decanus would you be able to unblock this merge? Thanks 😄

decanus (approved):

PR #118, opened 2020-05-19
Update PoW requirements

cammellos:
Updated the PoW requirements that we'd like to change in order to send media types.

README: Ignore the first commit in the review, it just moves the file to draft.

### DIFF
--- a/docs/draft/3-whisper-usage.md
+++ b/docs/draft/3-whisper-usage.md
@@ -1,16 +1,16 @@
 ---
 permalink: /spec/3
-parent: Stable specs
+parent: Draft specs
 title: 3/WHISPER-USAGE
 ---
 
 # 3/WHISPER-USAGE
 
-> Version: 0.2
+> Version: 0.3
 >
-> Status: Stable
+> Status: Draft
 >
-> Authors: Adam Babik <adam@status.im>, Corey Petty <corey@status.im>, Oskar Thorén <oskar@status.im> (alphabetical order)
+> Authors: Adam Babik <adam@status.im>, Andrea Maria Piana <andreap@status.im>, Corey Petty <corey@status.im>, Oskar Thorén <oskar@status.im> (alphabetical order)
 
 - [Status Whisper Usage Specification](#status-whisper-usage-specification)
   - [Abstract](#abstract)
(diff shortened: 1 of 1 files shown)


docs/draft/3-whisper-usage.md
```diff
@@ -0,0 +1,358 @@
...
+  - [Terminology](#terminology)
+  - [Whisper packets](#whisper-packets)
+  - [Whisper node configuration](#whisper-node-configuration)
+  - [Handshake](#handshake)
+  - [Rate limiting](#rate-limiting)
+  - [Keys management](#keys-management)
+    - [Contact code topic](#contact-code-topic)
+    - [Partitioned topic](#partitioned-topic)
+    - [Public chats](#public-chats)
+    - [Generic discovery topic](#generic-discovery-topic)
```
decanus:
```suggestion
```

docs/draft/3-whisper-usage.md
```diff
@@ -0,0 +1,358 @@
...
+    - [One-to-one topic](#one-to-one-topic)
```
decanus:
```suggestion
```

decanus (changes requested):

docs/draft/3-whisper-usage.md
```diff
@@ -84,8 +85,9 @@ encryption properties to support asynchronous chat.
 If you want to run a Whisper node and receive messages from Status clients, it must be properly configured.
 
 Whisper's Proof Of Work algorithm is used to deter denial of service and various spam/flood attacks against the Whisper network. The sender of a message must perform some work which in this case means processing time. Because Status' main client is a mobile client, this easily leads to battery draining and poor performance of the app itself. Hence, all clients MUST use the following Whisper node settings:
-* proof-of-work requirement not larger than `0.002`
+* proof-of-work requirement not larger than `0.00001`
 * time-to-live not lower than `10` (in seconds)
+* any payload below `50000` bytes MUST be sent with a PoW Target of at least `0.002`, in order to maintain backward compatibility with version `1.3` and below
```
oskarth: Might want to be explicit with what 1.3 is a version of. In the context of this spec it seems odd (v0.2 vs 0.3). Perhaps related to main client spec?

cammellos: mmh, is actually the main client version, client is `1.3`, while specs are `0.3`, we don't have specs following the main client, which kind of makes sense as at times we don't release something that would impact specs (bugfixes for example).
 I can either add a reference to the main client, or actually change that to `0.2`, which also makes sense. Wdyt?

oskarth: I think keeping it internally consistent (referring to specs) makes more sense. In app changelog it could refer which spec it is related to etc.

https://github.com/status-im/specs/blob/master/docs/stable/1-client.md should probably be updated to a major version...

Samyoul: Maybe something like

```suggestion
* any payload below `50000` bytes MUST be sent with a PoW Target of at least `0.002`, in order to maintain backward compatibility with [Status app version `1.3`](https://github.com/status-im/status-react/releases/tag/untagged-079a6d98babfeaa3f8c0) and below
```

cammellos: added both, as I would think that's useful information for someone implementing a client

kdeme (approved):

Samyoul (approved):

oskarth (approved):

PR #119, opened 2020-05-19
spellcheck

decanus:


docs/stable/6-payloads.md
```diff
@@ -14,8 +14,8 @@ title: 6/PAYLOADS
 
 ## Abstract
 
-This specifications decribes how the payload of each message in Status looks
-like. It is primarly centered around chat and chat-related use cases.
+This specifications describes how the payload of each message in Status looks
+like. It is primarily centered around chat and chat-related use cases.
 
 The payloads aims be flexible enough to support messaging but also cases
```
Samyoul: I spotted this on while doing #114, if there are any other grammatical issues that you didn't cover already I'll address them here.

```suggestion
The payloads aim to be flexible enough to support messaging but also cases
```

corpetty (approved):
LGTM

docs/stable/6-payloads.md
decanus: @Samyoul let's not do a all grammar issues here. I don't want this to get too large

Samyoul:
What do you think about changing https://github.com/status-im/specs/pull/119/files#diff-53565848b9b7dcb282e708743790874aR338

to

```md
   * `symKeyID` - `String`: an ID of a symmetric key used to authenticate with the `Mailserver`, derived from Mailserver password.
```

PR #108, opened 2020-05-01
Dapp browser API usage spec

vitvly:

PR #135, opened 2020-06-09
Add js and webview to wordlist

vitvly:
Added js and webview (including case variants) to wordlist.

This is required in order for https://github.com/status-im/specs/pull/108 to pass the spell-check.


vitvly:
Oh, so stupid - added a wordlist change to the original PR. Sorry.

PR #137, opened 2020-06-17
Added public key compression specs

Samyoul:
## What has changed?

I've added detailed specs for the implementation of public key compression and decompression. The specifications detail the use of the following `multiformat` features:

- `multibase`
- `multicodec`
- `unsigned-varint`

`multiformat` is used to ensure that the implementation has as much flexibility and robustness as feasible.

## Why make the change?

The usage of key de/compression is outside the typical usage of public keys and requires a degree of background knowledge to correctly implement. The purpose of this specification change is to provide this needed background knowledge.

Please also see

- https://github.com/status-im/status-go/issues/1937
- https://github.com/status-im/status-go/pull/1990


docs/stable/2-account.md
```diff
@@ -213,12 +220,193 @@ All messages sent are encrypted with the public key of the destination and signe
 
 -->
 
+## Public Key Compression
+
+The node MUST provide functionality for the compression and decompression of public / chat keys.
```
oskarth: SHOULD? Otherwise previous versions would be incompatible with the account spec, which seems off.

oskarth:
Thanks for the spec PR! Is there any chance we could people from Core (or more broadly, Status client implementers like desktop) more involved in these spec reviews as well? Currently I only see @cammellos as part of Core on reviews, where the rest are part of Nimbus/Vac.

cc @andremedeiros @hesterbruikman @iurimatias

arnetheduck:
There's a few things I don't understand here:

* where do these keys expected to appear? ie are they a pure UX feature or do the keys appear on the wire?
* what are the BLS keys used for 
* why both G1/G2 field variants?

arnetheduck:
also, where did the BLS serializations come from? the process of standardizing them has not settled yet cc @mratsim

docs/stable/2-account.md
```diff
@@ -213,12 +220,193 @@ All messages sent are encrypted with the public key of the destination and signe
...
+
+```text
+zQ3shPyZJnxZK4Bwyx9QsaksNKDYTPmpwPvGSjMYVHoXHeEgB
+```
+
+### Key Encoding
+
+The node MUST use the [multiformats/multibase](https://github.com/multiformats/multibase) encoding protocol to interpret incoming key data and to return key data in a desired encoding.
+
+The node MUST support the following `multibase` encoding formats.
```
oskarth: why is this a MUST? what happens if they don't?

docs/stable/2-account.md
```diff
@@ -213,12 +220,193 @@ All messages sent are encrypted with the public key of the destination and signe
...
+base36upper,       K,    base36 [0-9a-z] case-insensitive - no padding,            draft
+base58btc,         z,    base58 bitcoin,                                           default
+base58flickr,      Z,    base58 flicker,                                           candidate
+base64,            m,    rfc4648 no padding,                                       default
+base64pad,         M,    rfc4648 with padding - MIME encoding,                     candidate
+base64url,         u,    rfc4648 no padding,                                       default
+base64urlpad,      U,    rfc4648 with padding,                                     default
+```
+
+**Note** this specification RECOMMENDs that implementations extends the standard `multibase` protocol to parse strings prepended with `0x` as `f` hexadecimal encoded bytes.
```
oskarth: how come?

docs/stable/2-account.md
```diff
@@ -213,12 +220,193 @@ All messages sent are encrypted with the public key of the destination and signe
...
+This specification RECOMMENDs that for compression purposes the consuming service of the node uses a compact encoding type, such as base64 or base58 to allow for as short representations of the key as possible.
+
+### Public Key Types
+
+The node MUST support the [multiformats/multicodec](https://github.com/multiformats/multicodec) key type identifiers for the following public key types.
+
+| Name               | Tag | Code   | Description                          |
+| ------------------ | --- | ------ | ------------------------------------ |
+| `secp256k1-pub`    | key | `0xe7` | Secp256k1 public key                 |
+| `bls12_381-g1-pub` | key | `0xea` | BLS12-381 public key in the G1 field |
```
oskarth: what's the reason for BLS being in here? Without knowing too much about details, my understanding that this is used primarily for signature aggregation, e.g. in Eth2

docs/stable/2-account.md
```diff
@@ -213,12 +220,193 @@ All messages sent are encrypted with the public key of the destination and signe
...
 
 -
 
 ## Changelog
 
+### Version 0.4
+
+Released // TODO
+
+- Added details of public key compression and decompression
```
oskarth: The way the spec is currently writing this is currently a breaking change

oskarth (changes requested):
Some questions

docs/stable/2-account.md
Samyoul: Yes, you make a very good point.

Samyoul:
@arnetheduck 

>- where do these keys expected to appear? ie are they a pure UX feature or do the keys appear on the wire?

Yes, this functionality is (currently) purely for UX purposes, see my comment here https://github.com/status-im/status-go/issues/1937#issuecomment-638987966 and the issue on https://github.com/status-im/status-react/issues/10325

>- what are the BLS keys used for
>- why both G1/G2 field variants?

In my implementation I added BLS key handling for the sake of providing functionality that was simple to implement and potentially useful in the future. But on reflection I can't justify the mandatory use of BLS keys in the spec. I'll make them a RECOMMENDATION in the case that there are UX reasons for using BLS keys.

>also, where did the BLS serializations come from? the process of standardizing them has not settled yet

The BLS serialisation came from the zkcrypto spec https://github.com/zkcrypto/pairing/tree/master/src/bls12_381#serialization

docs/stable/2-account.md
Samyoul: I'm going rewrite the language to make public key de/compression implementation a SHOULD and make all MUSTs conditional on if de/compression functionality is being implemented.

docs/stable/2-account.md
Samyoul: I'm going to removed BLS from a MUST support key type and make them a RECOMMENDATION dependant on any UX requirements of these key types.

docs/stable/2-account.md
Samyoul: The next sentence explains the reason

>Implementing this recommendation will allow the node to correctly interpret traditionally identified hexadecimal strings (e.g. `0x1337c0de`).

Unless your point is, why allow users the option to use a non-standard identifier when the key data needs to be modified to support `multicodec` key identifiers anyway?

It would be for purely aesthetic / UX reasons. A hex encoded key is more easily recognised as "Ethereumy" by a user if it is prepended by `0x` than if it was prepended with `f`.

docs/stable/2-account.md
Samyoul: Hmm, I actually can't come up with a good justification for why this is a MUST. I will downgrade to a SHOULD.

docs/stable/2-account.md
oskarth: Ah right, I read it backwards, as `f` to `0x`. Makes sense

docs/stable/2-account.md
oskarth: how come they are recommended?

oskarth: if there's no specific use for it now I'd leave it out. If there MAY be use in the future it can be included as a MAY. If it SHOULD there should be some rationale imo, don't currently see why we'd go away from Secp256k1 for these types of keys (not Eth2 related operations)

arnetheduck: indeed - this standard is still in flux with the IETF - no point in including things just in case - can update the spec when and if they become standards

arnetheduck:
> In my implementation I added BLS key handling for the sake of providing functionality that was simple to implement and potentially useful in the future

YAGNI / it's going to be obsolete by the time of that future

docs/stable/2-account.md
```diff
@@ -213,12 +220,193 @@ All messages sent are encrypted with the public key of the destination and signe
 
 -->
 
+## Public Key Compression
```
arnetheduck: this is serialization - it's not compression, as in zip files. when serializing, you have the option to do so in a compressed format which replaces one of the two coordinates with a flag - you don't decompress it either, you deserialize or parse it

docs/stable/2-account.md
Samyoul: Ok, this is reasonable. I will remove mention of BLS usage

docs/stable/2-account.md
```diff
@@ -213,12 +220,193 @@ All messages sent are encrypted with the public key of the destination and signe
...
+| `bls12_381-g2-pub` | key | `0xeb` | BLS12-381 public key in the G2 field |
+
+For a public key to be identifiable to the node the public key data MUST be prepended with the relevant [multiformats/unsigned-varint](https://github.com/multiformats/unsigned-varint) formatted code.
+
+*Example:*
+
+Below is a representation of an uncompressed secp256k1 public key.
+
+```go
+[]byte{
```
arnetheduck: hm, are we now using `golang` as lingua franca in specs? the extra `[]byte{` here, do we define what it is somewhere?

Samyoul: I've adjusted this. Should we presume the reader understands that the bytes are encoded in hex and are ordered left to right, top to bottom?

docs/stable/2-account.md
Samyoul: Most libraries call this functionality "compress" and "decompress", or "from_compressed", "to_compressed", "from_decompressed" and "to_decompressed". The common parlance for this functionality is compress and decompress.

To be technically accurate I can add to the Terms Glossary that the terms "public key compression" and "public key decompression" are colloquial terms for the technical terms "public key serialisation" and "public key deserialisation".

The target reader and beneficiary of this document is much more likely to search for and understand the terms compress and decompress, so my opinion is that we keep the idiomatic phraseology in the main body of the document for this reason.

docs/stable/2-account.md
Samyoul: Done

docs/stable/2-account.md
```diff
@@ -213,12 +222,183 @@ All messages sent are encrypted with the public key of the destination and signe
...
+e7 | 01 | 02
+26 | 1c | 55 | 67 | 5e | 55 | ff | 25
+ed | b5 | 0b | 34 | 5c | fb | 3a | 3f
+35 | f6 | 07 | 12 | d2 | 51 | cb | aa
+ab | 97 | bd | 50 | 05 | 4c | 6e | bc
+```
+
+#### Decompression Example
+
+For the user the decompression process is exactly the same as compression with the exception that the user MUST provide a compressed public key for decompression. Else the decompression algorithm will fail.
```
decanus:
```suggestion
For the user, the decompression process is exactly the same as compression with the exception that the user MUST provide a compressed public key for decompression. Else the decompression algorithm will fail.
```

decanus (changes requested):

docs/stable/2-account.md
```diff
@@ -39,6 +39,13 @@ This specification explains what Status account is, and how a node establishes t
         - [Identicon](#identicon)
         - [3 word pseudonym / Whisper/Waku key fingerprint](#3-word-pseudonym--whisperwaku-key-fingerprint)
         - [ENS name](#ens-name)
+- [Public Key Compression](#public-key-compression)
```
oskarth: I'm missing some rationale for this feature. I know the gist of it but it'd be good to have it documented. Where exactly is it used etc?

oskarth (approved):
generally lgtm but some more clear rationale / where it is used would be useful

docs/stable/2-account.md
Samyoul: On reflection I've decided to implement this differently, I've taken your suggestion on board and now use the "serialize" terminology, with only a minor sentence detailing the idiomatic version of the technically correct phrases.

docs/stable/2-account.md
Samyoul: I've added a new subsection detailing where exactly this functionality would be used and why this functionality would be used. https://github.com/status-im/specs/pull/137/commits/d4cc96c42baf784c5890b4fdd4ebd5e4badfcb3e

Samyoul:
@cammellos and @gravityblast would you mind giving me your thoughts on this spec submission?

@decanus would you mind re-reviewing the changes made to this submission?

@oskarth do you feel that the rationale given is sufficient and gives sufficiently specific information about the places that would use key "compression"?

Thank you

oskarth:
@Samyoul seems good to me, thanks!

cammellos (approved):
I am still not completely sold on the why, but I won't rehash the conversation :) 
But otherwise  looks good to me

Samyoul:
@decanus would you mind unblocking the merge of this PR? Thank you. 🙂

decanus (approved):

PR #138, opened 2020-06-18
Add audio messages

cammellos:
adds audio messages, same method as images.

### DIFF
--- a/docs/draft/6-payloads.md
+++ b/docs/draft/6-payloads.md
@@ -6,7 +6,7 @@ title: 6/PAYLOADS
 
 # 6/PAYLOADS
 
-> Version: 0.3
+> Version: 0.4
 >
 > Status: Draft
 >
@@ -35,6 +35,7 @@ as various clients created using different technologies.
      - [Content types](#content-types)
        - [Sticker content type](#sticker-content-type)
        - [Image content type](#image-content-type)
+       - [Audio content type](#audio-content-type)
      - [Message types](#message-types)
      - [Clock vs Timestamp and message ordering](#clock-vs-timestamp-and-message-ordering)
      - [Chats](#chats)
@@ -50,6 +51,8 @@ as various clients created using different technologies.
    - [MembershipUpdateMessage and MembershipUpdateEvent](#membershipupdatemessage-and-membershipupdateevent)
  - [Upgradability](#upgradability)
  - [Security Considerations](#security-considerations)
+ - [Changelog](#changelog)
+ - [Copyright](#copyright)
 
 ## Introduction
 
@@ -57,7 +60,7 @@ This document describes the payload format and some special considerations.
 
 ## Payload wrapper
 
-All payloads are wrapped in a [protobuf record](https://developers.google.com/protocol-buffers/)
+The node wraps all payloads in a [protobuf record](https://developers.google.com/protocol-buffers/)
 record:
 
 ```protobuf
@@ -89,7 +92,7 @@ The protobuf description is:
 message ChatMessage {
   // Lamport timestamp of the chat message
   uint64 clock = 1;
-  // Unix timestamps in milliseconds, currently not used as we use whisper as more reliable, but here
+  // Unix timestamps in milliseconds, currently not used as we use Whisper/Waku as more reliable, but here
   // so that we don't rely on it
   uint64 timestamp = 2;
   // Text of the message
@@ -98,7 +101,9 @@ message ChatMessage {
   string response_to = 4;
   // Ens name of the sender
   string ens_name = 5;
-  // Chat id is the ID of the chat
+  // Chat id, this field is symmetric for public-chats and private group chats,
+  // but asymmetric in case of one-to-ones, as the sender will use the chat-id
+  // of the received, while the receiver will use the chat-id of the sender.
   string chat_id = 6;
 
   // The type of message (public/one-to-one/private-group-chat)
@@ -109,6 +114,7 @@ message ChatMessage {
   oneof payload {
     StickerMessage sticker = 9;
     ImageMessage image = 10;
+    AudioMessage audio = 11;
   }
 
   enum MessageType {
@@ -128,6 +134,7 @@ message ChatMessage {
     // Only local
     SYSTEM_MESSAGE_CONTENT_PRIVATE_GROUP = 6;
     IMAGE = 7;
+    AUDIO = 8;
   }
 }
 ```
@@ -144,7 +151,7 @@ message ChatMessage {
 | 6 | chat_id | `string` | The local ID of the chat the message is sent to |
 | 7 | message_type | `MessageType` | The type of message, different for one-to-one, public or group chats |
 | 8 | content_type | `ContentType` | The type of the content of the message | 
-| 9 | payload | `Sticker|Image|nil` | The payload of the message based on the content type |
+| 9 | payload | `Sticker|Image|Audio|nil` | The payload of the message based on the content type |
 
 #### Content types
 
@@ -159,6 +166,16 @@ There are other content types that MAY be implemented by the client:
 * `EMOJI`
 * `TRANSACTION_COMMAND`
 * `IMAGE`
+* `AUDIO`
+
+##### Mentions 
+
+A mention MUST be represented as a string with the `@0xpk` format, where `pk` is the public key of the [user account](https://specs.status.im/spec/2) to be mentioned, within the `text` field of a message with content_type `TEXT_PLAIN`.
+A message MAY contain more than one mention.
+This specification RECOMMENDs that the application does not require the user to enter the entire pk. 
+This specification RECOMMENDs that the application allows the user to create a mention by typing @ followed by the related ENS or 3-word pseudonym.
+This specification RECOMMENDs that the application provides the user auto-completion functionality to create a mention.
+For better user experience, the client SHOULD display a known [ens name or the 3-word pseudonym corresponding to the key](https://specs.status.im/spec/2#contact-verification) instead of the `pk`.
 
 ##### Sticker content type
 
@@ -175,7 +192,14 @@ message StickerMessage {
 ##### Image content type
 
 A `ChatMessage` with `IMAGE` `Content/Type` MUST also specify the `payload` of the image
-and the `type`
+and the `type`.
+
+Clients MUST sanitize the payload before accessing its content, in particular: 
+- Clients MUST choose a secure decoder
+- Clients SHOULD strip metadata if present without parsing/decoding it
+- Clients SHOULD NOT add metadata/exif when sending an image file for privacy and security reasons
+- Clients MUST make sure that the transport layer constraints the size of the payload to limit they are able to handle securely
+
 
 ```protobuf
 message ImageMessage {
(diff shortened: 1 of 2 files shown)


docs/draft/6-payloads.md
```diff
@@ -35,6 +35,7 @@ as various clients created using different technologies.
      - [Content types](#content-types)
```
oskarth, line 35: Changelog

docs/draft/6-payloads.md
```diff
@@ -191,6 +195,22 @@ message ImageMessage {
...
 ```
 
+##### Audio content type
+
+A `ChatMessage` with `AUDIO` `Content/Type` MUST also specify the `payload` of the audio
+and the `type`
+
+```protobuf
+message AudioMessage {
+  bytes payload = 1;
```
oskarth, line 231: Any max size?

docs/draft/6-payloads.md
```diff
@@ -191,6 +195,22 @@ message ImageMessage {
...
+  AudioType type = 2;
+  enum AudioType {
+    UNKNOWN_AUDIO_TYPE = 0;
+    AAC = 1;
```
oskarth, line 236: Any security concerns?

@rajeevgopalakrishna

oskarth (commented):
lgtm, few comments

docs/draft/6-payloads.md
rajeevgopalakrishna: Potential concerns:
1. Spam and network congestion given the size of audio msgs? Should the TTL be different for such msgs?
2. Any upper limit on the size of audio msgs?
3. Any impact on the PoW for audio msgs?
4. Is AAC the only supported format? What happens if a corrupt AAC file is used?

Anything else @status-im/security-internal?

docs/draft/6-payloads.md
cammellos: No, same as images, we don't limit size at this layer, the max size that is currently valid for any waku message (1MB) still stands.
If we want to allow users to send more data, it will not be handled at this layer but closer to the transport (by splitting the whole protobuf into chunks)

docs/draft/6-payloads.md
cammellos: Thanks for the points.

1) Currently there's no difference between an audio file or a text message in terms of bandwidth. Anyone can fill up a text file with 0s and 1s and send it through the network and it will be received by the users. This change in the specs does not change the maximum size limit.
2) We are not going to limit the size here, currently any envelope in waku is capped to 1MB, so it's implicitly capped for now. If there's a need to send larger files we can address that at a different layer, by splitting the whole protobuf (encoded) into chunks.
3) No, PoW would not be impacted, we lowered it for images already, but effectively this allows only to send a different type of content, but does not alter the quantity of data transmitted
4) I have added AMR as well, the protocol won't be checking that the file is properly formatted, I would leave that responsibility to the actual player, so that would be the clients responsibility to ensure that corrupted files are handled appropriately.

docs/draft/6-payloads.md
oskarth: right, forgot we dealt with that above already

docs/draft/6-payloads.md
```diff
@@ -159,6 +167,16 @@ There are other content types that MAY be implemented by the client:
...
+* `AUDIO`
+
+##### Mentions 
+
+A mention MUST be represented as a string with the `@0xpk` format, where `pk` is the public key of the [user account](https://specs.status.im/spec/2) to be mentioned, within the `text` field of a message with content_type `TEXT_PLAIN`.
+A message MAY contain more than one mention.
+This specification RECOMMENDs that the application does not require the user to enter the entire pk. 
+This specification RECOMMENDs that the application allows the user to create a mention by typing @ followed by the related ENS or 3-word pseudonym.
+This specification RECOMMENDs that the application provides the user auto-completion functionality to create a mention.
+For better user experience, the client SHOULD display a known [ens name or the 3-word pseudonym corresponding to the key](https://specs.status.im/spec/2#contact-verification) instead of the `pk`.
```
oskarth, line 178: Might be outside of PR but it shows up in diff here - can probably link to markdown directly

docs/draft/6-payloads.md
```diff
@@ -159,6 +167,16 @@ There are other content types that MAY be implemented by the client:
 * `EMOJI`
 * `TRANSACTION_COMMAND`
 * `IMAGE`
+* `AUDIO`
+
+##### Mentions 
+
+A mention MUST be represented as a string with the `@0xpk` format, where `pk` is the public key of the [user account](https://specs.status.im/spec/2) to be mentioned, within the `text` field of a message with content_type `TEXT_PLAIN`.
```
oskarth, line 173: Same as below

docs/draft/6-payloads.md
```diff
@@ -191,11 +209,28 @@ message ImageMessage {
...
+    AAC = 1;
+    AMR = 2;
+  }
+}
+```
+
 #### Message types
 
-A node requires message types to decide how to encrypt a particular message and what metadata needs to be attached when passing a message to the transport layer. For more on this, see [3/WHISPER-USAGE](./../stable/3-whisper-usage.md) and [10/WAKU-USAGE](./../stable/10-waku-usage.md).
+A node requires message types to decide how to encrypt a particular message and what metadata needs to be attached when passing a message to the transport layer. For more on this, see [3/WHISPER-USAGE](3-whisper-usage.md) and [10/WAKU-USAGE](10-waku-usage.md).
```
oskarth: what's going on with the remove of stable path?

oskarth (commented):
just some quick qs, somewhat unrelated to this pr, but in general lgtm

docs/draft/6-payloads.md
cammellos: my bad, I have integrated the changes from `stable`

docs/draft/6-payloads.md
0kok0: Could you add the assumption that the client is expected to sanitize the file from the message? I think its good to express that in the spec.

cammellos: Sure, would you use exactly the same wording, something like:

`clients SHOULD sanitize the payload before accessing its content`

or rather there's some better way to phrase this?

docs/draft/6-payloads.md
```diff
@@ -337,7 +372,7 @@ message PairInstallation {
 ### MembershipUpdateMessage and MembershipUpdateEvent
 
 `MembershipUpdateEvent` is a message used to propagate information about group membership changes in a group chat.
-The details are in the [Group chats specs](./7-group-chat.md).
+The details are in the [Group chats specs](./../draft/7-group-chat.md).
```
Samyoul: I'm sure you are already aware, but this link too was changed.

Samyoul (approved):
Looks good, just got them annoying link tweaks AFAICS

docs/draft/6-payloads.md
```diff
@@ -98,7 +101,10 @@ message ChatMessage {
   string response_to = 4;
   // Ens name of the sender
   string ens_name = 5;
-  // Chat id is the ID of the chat
+  // Chat id, this field is symmetric for public-chats and private group chats,
+  // but asymmetric in case of one-to-ones, as the sender will use the chat-id
+  // of the received, while the receiver will use the chat-id of the sender.
+  // Probably should be the concatenation of sender-pk & receiver-pk in alphabetical order
```
decanus: "Probably" ? that doesn't sound very technical.

docs/draft/6-payloads.md
```diff
@@ -191,11 +209,28 @@ message ImageMessage {
 }
 ```
 
+##### Audio content type
+
+A `ChatMessage` with `AUDIO` `Content/Type` MUST also specify the `payload` of the audio
+and the `type`
```
decanus:
```suggestion
and the `type`.
```

decanus (changes requested):

docs/draft/6-payloads.md
0kok0: I think that's a MUST case. I'd suggest:

Clients MUST sanitize the payload before accessing its content. In particular, clients MUST verify that the payload size is constrained and a secure decoder is chosen on the receivers end.

oskarth (approved):

decanus (approved):

docs/draft/6-payloads.md
```diff
@@ -350,6 +385,21 @@ There are two ways to upgrade the protocol without breaking compatibility:
```
0kok0, line 398:
```suggestion
Clients MUST sanitize the payload before accessing its content, in particular: 
- Clients MUST choose a secure decoder
- Clients SHOULD strip metadata if present without parsing/decoding it
- Clients SHOULD NOT add metadata/exif when sending an audio file for privacy and security reasons
- Clients MUST make sure that the transport layer constraints the size of the payload to limit they are able to handle securely
```

0kok0 (changes requested):

0kok0:
Added some security considerations, similar considerations apply as for [image metadata](https://github.com/status-im/status-react/issues/10769)
Associated attack vector got exploited recently:
 https://blog.malwarebytes.com/threat-analysis/2020/06/web-skimmer-hides-within-exif-metadata-exfiltrates-credit-cards-via-image-files/

cammellos:
@0kok0 integrated the changes, ready to be reviewed again, thanks!

docs/draft/6-payloads.md
```diff
@@ -175,7 +192,14 @@ message StickerMessage {
 ##### Image content type
 
 A `ChatMessage` with `IMAGE` `Content/Type` MUST also specify the `payload` of the image
-and the `type`
+and the `type`.
+
+Clients MUST sanitize the payload before accessing its content, in particular: 
+- Clients MUST choose a secure decoder
+- Clients SHOULD strip metadata if present without parsing/decoding it
+- Clients SHOULD NOT add metadata/exif when sending an image file for privacy and security reasons
```
andremedeiros, line 200: Should this be `MUST NOT` instead?

docs/draft/6-payloads.md
```diff
@@ -191,11 +215,35 @@ message ImageMessage {
...
 
+##### Audio content type
+
+A `ChatMessage` with `AUDIO` `Content/Type` MUST also specify the `payload` of the audio,
+the `type` and the duration in milliseconds (`duration_ms`).
+
+Clients MUST sanitize the payload before accessing its content, in particular: 
+- Clients MUST choose a secure decoder
+- Clients SHOULD strip metadata if present without parsing/decoding it
+- Clients SHOULD NOT add metadata/exif when sending an audio file for privacy and security reasons
```
andremedeiros, line 226: Same as above.

docs/draft/6-payloads.md
0kok0: I'd sure prefer a must not, I wanted to leave some space for features that rely on optional EXIF

docs/draft/6-payloads.md
0kok0: Better add a note that mention resolution via ENS should be limited in the client or could be exploited for reflected DOS especially in large channels. 
https://github.com/status-im/specs/issues/144
https://www.cloudflare.com/learning/ddos/dns-amplification-ddos-attack/

0kok0 (approved):

PR #139, opened 2020-06-22
Push notification server specs

cammellos:
Specs for the push notification server.
Since this is a quite a large document, I would do review in rounds, 

First, please provide feedback  on the actual protocol, methods etc

After the first phase we can move onto formatting/sections etc and take it from there.


docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,672 @@
...
+
+- [Push Notification Server](#16push-notification-server)
+  - [Reason](#reason)
+  - [Requirements](#requirements)
+  - [Registering with the push notification service](#registering-with-the-push-notification-server)
+  - [Re-registering with the push notification service](#re-registering-with-the-push-notification-server)
+  - [Changing options](#changing-options)
+  - [Unregistering from push notifications](#unregistering-from-push-notifications)
+  - [Advertising a push notification server](#advertising-a-push-notification-server)
+  - [Discovering a push notification server](#discovering-a-push-notificatino-server)
```
decanus:
```suggestion
  - [Discovering a push notification server](#discovering-a-push-notification-server)
```

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,672 @@
...
+  - [Querying the push notification service](#querying-the-push-notification-server)
+  - [Sending a push notification](#sending-a-push-notification)
+  - [Flow](#flow)
+    - [Registration process](#registration-process)
+    - [Sending a notification](#sending-a-notification)
+    - [Receiving a notification](#receiving-a-notification)
```
decanus:
```suggestion
    - [Receiving a push notification](#receiving-a-push-notification)
```

decanus (changes requested):

3esmit:
I think that is possible to implement push notifications without compromising privacy. It's not necessary to inform user public_key to the server, client can inform anything it wants and contacts can use an additional message just for notifications for that particular user, assuming that the sender knows the notification topic.

---

Current proposal. public key as notification topics:
- Notification Topics/Channels: User public key
- Users subscribe to the notification service, informing the list of topics they listen.
- Notification service is able to identify when user receives a message. 

Advantages: Simple
Disadvantages: not private, vulnerable to notification flood

---

Proposal 1, ephemeral notification topics (contacts only):
- Notification Topics/Channels: When someone adds other in contact,  they send a protocol message informing what is the topic they listen for firing "a notification" for that user. 
- Users subscribe to the notification service, informing the list of topics they listen.
- The notification topic does not contain the message being sent, just the notification request.
- The notification topic changes from time to time, and might use asymmetric or symmetric encryption.
- Users only receive notifications from users in their contact list. 
- "Handshakes" are excluded from the push notification, as they happen while the Status App is open, while users add another user as contact. 

Advantages: Safe against flooding, very private
Disadvantages: Complex, less scalable, 

---

Proposal 2, public fixed notification topics:
- Notification Topics/Channels: A public channel for each key can be derived by hashing the public key. 
- Users subscribe to a "hash of their public key"
- Users are able to derive other's notification topic.

Advantages: Simple, somewhat private
Disadvantages: Vulnerable to notification flood

---

Proposal 3, derived notifications topics:
- Notification Topics/Channels: A public channel for each contact pair, derived from hashing receiver pubkey and sender pubkey
- When adding a contact, users subscribe to everty "hash of their public key+contact private key",

Advantages: Simple, somewhat private
Disadvantages: less scalable

oskarth:
@decanus I think OP explicitly asked for non formatting PR comments at this stage. Do you have any feedback on the actual meat of the spec?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,672 @@
...
+    - [PushNotificationAcknowledgement](#pushnotificationacknowledgement)
+  - [Partitioned topic](#partitioned-topic)
+  - [FAQ](#faq)
+  - [Changelog](#changelog)
+    - [Version 0.1](#version-01)
+  - [Copyright](#copyright)
+
+## Reason
+
+Push notifications for IOS devices and for some android device can only be
```
oskarth: Comment for second round: is this strictly true? in practice this is the conclusion. I'd perhaps elaborate a bit more remote and local options here and why remote for iOS makes sense, and that this implies using the APN service

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,672 @@
...
+background service.
+Therefore Status provides a set of Push notification services that can be used
+to achieve this functionality.
+
+Because this can't be safely implemented in a privacy preserving manner, clients
+MUST be given an option not to enable push notifications and not to send push notifications.
+
+## Requirements
+
+Status needs to run their own [gorush](https://github.com/appleboy/gorush)
```
oskarth: Second round: the party releasing the app. Which could be Status Gmbh (or PTE LTD?), or some other party.

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,672 @@
...
+public key of the node, encrypted with this key.
+
+The content of the message MUST contain the following [protobuf record](https://developers.google.com/protocol-buffers/):
+
+```protobuf
+
+message PushNotificationOptions {
+  boolean enabled = 1;
+  repeated string allowed_user_list = 2;
+  repeated string block_user_list = 3;
```
oskarth, line 116: Is this supposed to be appending? Could this get out of hand in terms of the number of users allowed/blocked? (I suppose that can be taken care of later)

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,672 @@
...
+    FIREBASE_TOKEN = 2;
+  }
+  TokenType token_type = 1;
+  string token = 2;
+  string installation_id = 3;
+  PushNotificationPreferences preferences = 4;
+}
+
+message PushNotificationPreferences {
+  repeated DeviceToken device_tokens = 1;
```
oskarth, line 133: Outside of DeviceToken, is the unique identifier here DeviceToken+ID? How do we deal with the 1-device-N-user case here?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,672 @@
...
+
+If the `version` is equal or less than the currently stored version, a response MUST
+be sent with `error` set to `VERSION_MISMATCH`.
+
+If any other error occurs the `error` should be set to `INTERNAL_ERROR`.
+
+If the request is successful, an access token MUST be generated and sent back to the client
+with `success` set to `true`.
+
+It is RECOMMENDED to use a randomly generated string of at least 36 characters.
```
oskarth: for which field?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,672 @@
...
+Otherwise a response MUST be sent with `success` set to `false`.
+
+`request_id` should be set to the `SHA3` of the signature sent by the client.
+
+The request MUST be sent on the [partitioned topic][./2-account.md] of the sender.
+
+The payload of the response is:
+
+```protobuf
+message PushNotificationPreferencesAcknowledgement {
```
oskarth: is this an Ack or more of a Reply? If a generic Ack, any reason we couldn't use data sync for this? Reading fields, it seems more like a Reply to me.

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,672 @@
...
+
+# 16/PUSH-NOTIFICATION-SERVER
+
+> Version: 0.1
+>
+> Status: Raw
+>
+> Authors: Andrea Maria Piana <andreap@status.im>
+
+- [Push Notification Server](#16push-notification-server)
```
oskarth, line 15: I wonder if it makes sense to have a basic components section, so it is clear what the pieces are. E.g.

- Central PN server
- PN service
- Alice Device-ID session
- Bob Device-ID session

Can e..g be a graphviz diagram to showcase how registering works, etc.  E.g. https://specs.vac.dev/specs/remote-log.html#flow

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,672 @@
...
+MUST be given an option not to enable push notifications and not to send push notifications.
+
+## Requirements
+
+Status needs to run their own [gorush](https://github.com/appleboy/gorush) 
+publicly accessible server for sending the actual notification. This cannot be community run as it requires sharing private credentials that only status has access to.
+
+## Registering with the push notification service
+
+A client MAY register with a Push Notification service of their choice.
```
oskarth: can they register / get replies from multiple? (partially answered below)

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,672 @@
...
+
+If `success` is `false`:
+
+- If `MALFORMED_MESSAGE` is returned, the request SHOULD not be retried without ensuring 
+  that is correctly formed.
+- If `INTERNAL_ERROR` is returned, the request MAY be retried, but the client MUST be 
+  backoff exponentially
+- If `VERSION_MISMATCH` is returned, the client SHOULD ensure that the remote version returned in `preferences` is integrated, and the version should be incremented and the request MAY be sent again.
+
+A client MAY register with multiple Push Notification Servers in order to increase availability.
```
oskarth, line 211: I see above is answered here

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,672 @@
...
+Once re-registered, a client SHOULD advertise the changes.
+
+
+## Changing options
+
+This is handled in exactly the same way as re-registering above.
+
+## Unregistering from push notifications
+
+To unregister a client MUST send a `PushNotificationRegister` request as described
```
oskarth, line 237: Should this be explicit in terms of 'when a user logs out' the client should deregister?

I suppose that is part of a the main client spec, not here

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,672 @@
...
+above with `unregister` in `PushNotificationPreferences` set to `true`, or removing
+their device information.
+
+The server MUST remove all data about this user if `unregistering` is `true`,
+apart from the `hash` of the public key and the `version` of the last options,
+in order to make sure that old messages are not processed.
+
+## Advertising a push notification server
+
+Each user registered with one or more push notification servers SHOULD
```
oskarth, line 250: Just to make sure I understand correctly: Is the idea then that each user who wants to send PNs to that user would pick this up and store it locally?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,672 @@
...
+
+Where `message` is the encrypted payload of the message and `chat_id` is the
+`SHA3` of the `chat_id`.
+`message_id` is the id of the message (link)
+
+If multiple server are available for a given push notification, only one notification
+MUST be sent.
+In such cases `ack_required` MAY bet set and the sender SHOULD be listening
+on the topic derived from the first 4 bytes of `SHA3(ack_required)` and with
+waku encryption key of `ack_required`.
```
oskarth: elaborate on waku encryption key here

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,672 @@
...
+a bandwidth impact but they are made only when actually needed
+
+### What's the information disclosed?
+
+The data disclosed with each message sent by the client is above, but for a summary:
+
+When you register with a push notification service you may disclose:
+
+1) Your public key
+2) Which devices you have
```
oskarth, line 696: can you choose to disclose only some of those devices? for whatever reason

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,672 @@
...
+
+Effectively this is fairly revealing if the user has a whitelist implemented.
+Therefore sending notification should be optional.
+
+### What prevents a user from generating a random key and getting an access token and spamming?
+
+Nothing really, that's the same as the status app as a whole. the only mechanism that prevents
+this is using a whilelist as described above, but that implies disclosing your true identity to the push notification server.
+
+### Why not 0-knowledge proofs/quantum computing
```
oskarth, line 713: lol

oskarth (commented):
Initial comments, quite a lot to take in, in general looks good - nice job!

oskarth:
@status-im/security-internal probably need to have a look at this, added @rajeevgopalakrishna as that's the only username I know except for Corey :P

docs/raw/push-notification-server.md
oskarth: answered below

cammellos:
@3esmit 
We are not implementing notifications only for contacts, if that was the case it would be much simpler.
https://discuss.status.im/t/decentralized-push-notifications-for-ios/1718/3?u=cammellos
So 1 & 3 don't fit the requirements.

Proposal 2 is already incorporated. If you don't care about spam, you can advertise the token on your contact code topic (derived from your public key), and user can use the token to send you push notifications, without disclosing the identity. If that's the case, you can just use an ephemeral key to register with the push notification server, so you won't disclose your identity.

docs/raw/push-notification-server.md
cammellos: I lost a sentence somewhere :) it's for the access token

docs/raw/push-notification-server.md
cammellos: Reply is a more fitting name

docs/raw/push-notification-server.md
cammellos: Though is not on logout (otherwise you will stop receiving notifications), you will want to unregister if you don't want to receive notifications anymore, or don't want to use this server anymore.

docs/raw/push-notification-server.md
cammellos: Yes, we could have a discovery for which push notification server is used by a given user, that's simple:

Every PN server listen to topic(pk-of-user) and the client can query it and the pk of the server(s) is returned, but really offers no benefits at the moment compare to just querying a mailserver, so left it out for simplicity.

docs/raw/push-notification-server.md
cammellos: Yes, you can always decide not to register with a notification server, or partially register etc

docs/raw/push-notification-server.md
cammellos: well depends on your definition of push notification, if you want them to be instant (which is the common understanding), then yes, as we are not allowed to keep a connection open (only voip applications do), and the app would be brought to the foreground to check for messages at intervals, and only when apple thinks is energy efficient (on wifi, when the user switches application etc), and would not be brought back if force quit.

I can add a section to expand on the rationale.

docs/raw/push-notification-server.md
cammellos: This is straight replacing, given that this is a fairly rare message and we have a 1MB limit, I think we are safe in terms of size, if size will be a concern we can always implement a partial update.

docs/raw/push-notification-server.md
cammellos: > 1-device-N-user

Not sure I understand exactly the question, you mean multiple users using the same device? Which is essentially, how do we handle multiple accounts on a single installation?

If that's the case, is `installation_id`. Installation id is different for each generated account, so if you have 3 accounts on your device, you will have 3 different installation ids. this is passed to the push notification, so on opening the notification, we can already present the user with the right account and it can type the password.

hope I understood the question correctly.

docs/raw/push-notification-server.md
cammellos: I settled on Response

docs/raw/push-notification-server.md
cammellos: Yes

cammellos:
@oskarth replied with some clarifications and addressed the feedback, thanks

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,740 @@
...
+  - [Changelog](#changelog)
+    - [Version 0.1](#version-01)
+  - [Copyright](#copyright)
+
+## Reason
+
+Push notifications for IOS devices and for some android device can only be
+implemented relying on [APN service](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/APNSOverview.html#//apple_ref/doc/uid/TP40008194-CH8-SW1) for IOS or [Firebase](https://firebase.google.com/).
+
+This is useful for some Android devices that do not support background services or that often kill the
```
decanus:
```suggestion
This is useful for Android devices that do not support background services or that often kill the 
```

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,740 @@
...
+
+## Reason
+
+Push notifications for IOS devices and for some android device can only be
+implemented relying on [APN service](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/APNSOverview.html#//apple_ref/doc/uid/TP40008194-CH8-SW1) for IOS or [Firebase](https://firebase.google.com/).
+
+This is useful for some Android devices that do not support background services or that often kill the 
+background service.
+
+IOS only allows certain kind of applications to keep a connection open when in the
```
decanus:
```suggestion
iOS only allows certain kind of applications to keep a connection open when in the
```

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,740 @@
...
+background, VoIP for example, which current status client does not qualify for.
+
+Application on IOS can also request execution time when are in the [background]
```
decanus:
```suggestion
Applications on iOS can also request execution time when they are in the [background]
```

decanus (changes requested):

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,740 @@
...
+be sent with `error` set to `VERSION_MISMATCH`.
+
+If any other error occurs the `error` should be set to `INTERNAL_ERROR`.
+
+If the request is successful, an access token MUST be generated and sent back to the client
+with `success` set to `true`.
+
+It is RECOMMENDED to use a randomly generated string for `access_token` of at least 36 characters.
+
+Otherwise a response MUST be sent with `success` set to `false`.
```
decanus, line 169: Should this line not be moved up? it seems out of context with the above line of the access token.

decanus (changes requested):

decanus:
I think it probably makes sense to add what metadata is leaked etc to a `Security Considerations` section, you have this answered elsewhere but I think it probably belongs here. wdyt @oskarth

docs/raw/push-notification-server.md
oskarth: I agree with the conclusion, just that the decision space pruning should come first, i.e. can only do remote or local, local does'nt work due to X. If X were were to change in future (unlikely but hey), it is clear to reader that architecture could change.

docs/raw/push-notification-server.md
oskarth: Multiple accounts/users on one device, correct.

docs/raw/push-notification-server.md
oskarth: Hm, it's not clear to me what the correct design is here. If you logout as one user and login as a new user, you could argue that the first user shouldn't receive notifications anymore. To me that might be more intuitive, than getting PNs for some user user.

Use case: family sharing a phone, or some kind of kiosk mode where you shouldn't receive PNs for other suers.

oskarth (approved):
FYI: I'm not going to be able to review this until Monday again. I'd be happy to have this merged in a Raw state if everything else is taken care of. Obviously a lot of formatting etc stuff to be done, but I trust this will be identified in the second stage. The meat generally seems fine to me.

docs/raw/push-notification-server.md
cammellos: I double checked with UX and indeed is how they think it should be behave, until explicit logout you should receive notifications.
It is a bit of a UX concern, so not sure if it's needed to be specified here.
For now I will add it as a MAY and the implementation in status-react will reflect this behavior.

cammellos:
@decanus ready to be reviewed again

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,766 @@
...
+  PushNotificationPreferences preferences = 4;
+}
+
+message PushNotificationPreferences {
+  repeated DeviceToken device_tokens = 1;
+  uint version = 2;
+  boolean unregister = 3;
+}
+
+message PushNotificationRegister {
```
0kok0, line 138: How do we prevent this message to be intercepted and replayed to sign up for other servers? An attacker could do so to compromise privacy, spam notifications and if a cost is attached drain user funds

cammellos: Good question, so:

1) the message is sent to the server encrypted with their public key, so only the server is able to  access it. 

2) If replied to the same server the actions taken are idempotent.
3)  If you look at `PushNotificationRegister` definition, we `Sign(payload+publickeyoftheserver)` so that any server receiving this message will discard it if not for them.

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,766 @@
...
+
+If a token is not present in the latest advertisement for a user, the server 
+SHOULD be queried directly.
+
+To query a server a message:
+
+```protobuf
+message PushNotificationQuery {
+  repeated bytes public_keys = 1;
+}
```
0kok0, line 297: Same as above, we probably need a nonce or this message might be replayed to get access to the tokens

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,766 @@
...
+message PushNotificationRequest {
+  repeated PushNotification requests = 1;
+  bytes message = 2;
+  string message_id = 3;
+  string ack_required = 4;
+}
+```
+
+Where `message` is the encrypted payload of the message and `chat_id` is the
+`SHA3-256` of the `chat_id`.
```
0kok0, line 359: Why do we hash here? Is chatid random and the preimage space large enough?

docs/raw/push-notification-server.md
0kok0: Okay so with these assumptions replaying is only possible if an eavesdropper waits for a user to register and register.

To me the question remains if a server would still discard a notifications sign up if a payment is involved and potentially claimable with this message.

docs/raw/push-notification-server.md
cammellos: `ChatID` is random in some cases (group chats), and not random in some other (one to one) where is the public key of the user.
Preimage is large enough in both cases.

docs/raw/push-notification-server.md
cammellos: > Okay so with these assumptions replaying is only possible if an eavesdropper waits for a user to register and register.

Not sure I follow,
you can only reply this to the same server, as the signature would fail to verify and it would be discarded. 
Replying to the same server will not yield to any result, as the message would be discarded if already processed (it's basically idempotent), due to a `version` inside the message that is part of the hash to be signed, which effectively acts as a `nonce`.

Given the above, is this still replayable in any meaningful sense?

> To me the question remains if a server would still discard a notifications sign up if a payment is involved and potentially claimable with this message.

What do you mean exactly? You mean that someone could offer money to a malicious server to drop the message/reveal the message?

docs/raw/push-notification-server.md
cammellos: mmh, not sure, first the message is encrypted with the server pk, so to replay this you would need to have access to the private key of the server.

Say that this happen and the key is compromised,an attacker would yes be able to replay the same message, but it would not be able to listen to the reply, as it's encrypted with the pk of the user querying.

If the user has access to both, then it would not make sense to replay a message, as it can just craft a new one.

I mean, we can add a nonce, but in order to replay this effectively you would have to compromise two private keys (or one, the user querying, but at that point you can craft a new message, listen to replies etc), and this message does not change any state, so you would only replay it to get access to information.

Or you think there is still some reason to add a nonce?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,766 @@
...
+  - [Anonymous mode of operations](#anonymous-mode-of-operations)
+  - [Security considerations](#security-considerations)
+  - [FAQ](#faq)
+  - [Changelog](#changelog)
+    - [Version 0.1](#version-01)
+  - [Copyright](#copyright)
+
+## Reason
+
+Push notifications for IOS devices and for some android device can only be
```
decanus:
```suggestion
Push notifications for iOS devices and for some Android devices can only be
```

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,766 @@
...
+implemented relying on [APN service](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/APNSOverview.html#//apple_ref/doc/uid/TP40008194-CH8-SW1) for IOS or [Firebase](https://firebase.google.com/).
```
decanus:
```suggestion
implemented by relying on [APN service](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/APNSOverview.html#//apple_ref/doc/uid/TP40008194-CH8-SW1) for iOS or [Firebase](https://firebase.google.com/).
```

docs/raw/push-notification-server.md
0kok0: Could you elaborate on how the version could replace a nonce? I was thinking about case: user registers, user unregisters, attacker replays to register again.

For the case of malicious servers I was referring to the notifications as a service case (I didn't find specs for that but iirc it was somehow a longterm plan) , depending on the implementation, if the payment is part of the message payload a replay to a malicious server could be used to steal user funds

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,766 @@
...
+
+This is useful for Android devices that do not support background services or that often kill the 
+background service.
+
+iOS only allows certain kind of applications to keep a connection open when in the
+background, VoIP for example, which current status client does not qualify for.
+
+Applications on iOS can also request execution time when they are in the [background]
+(https://developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_background/updating_your_app_with_background_app_refresh)
```
Samyoul: Link doesn't show correctly

```suggestion
Applications on iOS can also request execution time when they are in the [background](https://developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_background/updating_your_app_with_background_app_refresh)
```

decanus (changes requested):

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,766 @@
...
```
decanus:
```suggestion
Applications on iOS can also request execution time when they are in the [background](https://developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_background/updating_your_app_with_background_app_refresh)
```

decanus (changes requested):

docs/raw/push-notification-server.md
Samyoul: Maybe make this sentence more clear that multiple PN services can be registered to.

```suggestion
A client MAY register with one or more Push Notification services of their choice.
```

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,766 @@
...
+A status client that wants to send push notifications
+
+## Registering with the push notification service
+
+A client MAY register with a Push Notification service of their choice.
+
+[//]: # (We can have a separate partitioned topic scheme, or individual based on PK)
+
+A message MUST be sent to the [partitioned topic](#partitioned-topic) for the
+public key of the node, encrypted with this key.
```
Samyoul: This is a special notification registration message, perhaps we should give this message a special dedicated name. Something like:

```suggestion
A `PNR message` (Push Notification Registration) MUST be sent to the [partitioned topic](#partitioned-topic) for the
public key of the node, encrypted with this key.
```

docs/raw/push-notification-server.md
cammellos: >  I was thinking about case: user registers, user unregisters, attacker replays to register again.

In this case it won't work, `version` is basically the same as an ethereum nonce (minus the fact that they need to be strictly incrementing by one), basically the server will only process requests that have a version strictly greater than the previous one.
So in your case, user will register with a version of `1`. They will unregister, as per spec this will need to have a version of `2`. The server will keep the last version (`2`) stored, as per spec:
```
The server MUST remove all data about this user if `unregistering` is `true`,
apart from the `hash` of the public key and the `version` of the last options,
in order to make sure that old messages are not processed.
```
If `1` (or `2`) is replayed, the server will ignore the message, as the `version` is not strictly greater than `2`

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,766 @@
...
+If the request is successful, an access token MUST be generated and sent back to the client
+with `success` set to `true`.
+
+Otherwise a response MUST be sent with `success` set to `false`.
+
+It is RECOMMENDED to use a randomly generated string for `access_token` of at least 36 characters.
+
+`request_id` should be set to the `SHA3-256` of the signature sent by the client.
+
+The request MUST be sent on the [partitioned topic][./2-account.md] of the sender.
```
Samyoul: Should this be `[partitioned topic](#partitioned-topic)`? I can't see any reference to partitioned topic in the account specs.

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,766 @@
...
+    VERSION_MISMATCH = 2;
+    UNSUPPORTED_TOKEN_TYPE = 3;
+    INTERNAL_ERROR = 4;
+  }
+}
+```
+
+[//]: (We can ratched and use a partitioned topic here)
+
+A client SHOULD listen for a response sent on their [partitioned topic][./2-account.md].
```
Samyoul: Same as above

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,766 @@
...
+  that is correctly formed.
+- If `INTERNAL_ERROR` is returned, the request MAY be retried, but the client MUST be 
+  backoff exponentially
+- If `VERSION_MISMATCH` is returned, the client SHOULD ensure that the remote version returned in `preferences` is integrated, and the version should be incremented and the request MAY be sent again.
+
+A client MAY register with multiple Push Notification Servers in order to increase availability.
+
+A client SHOULD make sure that all the notification services they registered with have the same information about their devices and tokens.
+
+If no response is returned the request SHOULD be considered failed and MAY be retried with the same server or a different one, but it MUST exponentially backoff after each trial.
```
Samyoul: >but it MUST exponentially backoff after each trial

Does "it" refer to a client?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,766 @@
...
+
+## Re-registering with the push notification server
+
+A client SHOULD re-register with the node if the APN or FIREBASE token changes.
+
+When re-registering a client SHOULD ensure that it has the most up-to-date
+`PushNotificationPreferences`, update the part relative to their `installation_id`
+if necessary, increment `version` and send a `PushNotificationRegister` as described above.
+
+Once re-registered, a client SHOULD advertise the changes.
```
Samyoul, line 228: Does the node need special behaviour for re-registering, or will the previously given registration details be overwritten and behave exactly the same as a standard registration?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,766 @@
...
+### Can someone else (i.e not status) run this?
+
+Push notification servers can be run by anyone. Gorush can be run by anyone I take,
+but we are in charge of the certificate, so they would not be able to notify status-clients.
+
+## Changelog
+
+### Version 0.1
+
+Released [May 22, 2020](https://github.com/status-im/specs/commit/664dd1c9df6ad409e4c007fefc8c8945b8d324e8)
```
Samyoul: Is this true? I thought that this spec was only drafted this month. Maybe just a copypasta from another file?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,766 @@
...
+Therefore sending notification should be optional.
+
+### What prevents a user from generating a random key and getting an access token and spamming?
+
+Nothing really, that's the same as the status app as a whole. the only mechanism that prevents
+this is using a white-list as described above, but that implies disclosing your true identity to the push notification server.
+
+### Why not 0-knowledge proofs/quantum computing
+
+We start simple, we can iterate
```
Samyoul, line 715: Beautifully succinct

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,766 @@
...
+review process. Point can be integrated, suggestion welcome.
+
+## Security considerations
+
+If no anonymous mode is used, when registering with a push notification service a client discloses:
+
+- Their public key
+- Which devices they want to receive notifications on
+
+They MAY disclose:
```
Samyoul: Does "They" refer to a client?

Samyoul (approved):
Congratulations on this intimidatingly impressive monster spec. Amazing work @cammellos

docs/raw/push-notification-server.md
oskarth: Fair, agre re deferring it.

@hesterbruikman IMO this UX decision should be revisited. It goes against expectations and also explicitly against previous PN designs I have made, and just sounds off to me. If you are logged off you should NOT receive PNs IMO, that's what that action does. I doubt it is how other IMs work either tbh.

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+>
+> Status: Raw
+>
+> Authors: Andrea Maria Piana <andreap@status.im>
+
+- [Push Notification Server](#16push-notification-server)
+  - [Reason](#reason)
+  - [Requirements](#requirements)
+  - [Components](#components)
+  - [Registering with the push notification service](#registering-with-the-push-notification-server)
```
decanus:
```suggestion
  - [Registering with the push notification service](#registering-with-the-push-notification-service)
```

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+  - [Anonymous mode of operations](#anonymous-mode-of-operations)
+  - [Security considerations](#security-considerations)
+  - [FAQ](#faq)
+  - [Changelog](#changelog)
+    - [Version 0.1](#version-01)
+  - [Copyright](#copyright)
+
+## Reason
+
+Push notifications for iOS devices and for some Android devices can only be
```
decanus:
```suggestion
Push notifications for iOS devices and some Android devices can only be
```

decanus (changes requested):

docs/raw/push-notification-server.md
cammellos: It should be `10-waku-usage.md#partitioned-topic` good catch

docs/raw/push-notification-server.md
cammellos: Yes, I'll use `clients` instead

docs/raw/push-notification-server.md
cammellos: No special behavior, the previously given registration details will be overwritten.

docs/raw/push-notification-server.md
cammellos: yes, changed to avoid possessive adjectives

docs/raw/push-notification-server.md
cammellos: No, that is indeed a copypasta, I will fill it in once ready to merge :)

cammellos:
@decanus addressed feedback, thanks

decanus (approved):

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+  - [Changelog](#changelog)
+    - [Version 0.1](#version-01)
+  - [Copyright](#copyright)
+
+## Reason
+
+Push notifications for iOS devices and some Android devices can only be
+implemented by relying on [APN service](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/APNSOverview.html#//apple_ref/doc/uid/TP40008194-CH8-SW1) for iOS or [Firebase](https://firebase.google.com/).
+
+This is useful for Android devices that do not support background services or that often kill the
```
rajeevgopalakrishna, line 58: It will help to mention which are these Android devices? Does it depend on the OEM or OS or something else? Also, my understanding is that our current Android "local" notifications depend on [foreground service](https://github.com/status-im/specs/blob/b31ce655b99580c3e5d10c81207ae829e561757e/docs/draft/notifications.md). So what does support for background services have to do with that? Perhaps they're related. Good to clarify.

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+
+## Reason
+
+Push notifications for iOS devices and some Android devices can only be
+implemented by relying on [APN service](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/APNSOverview.html#//apple_ref/doc/uid/TP40008194-CH8-SW1) for iOS or [Firebase](https://firebase.google.com/).
+
+This is useful for Android devices that do not support background services or that often kill the 
+background service.
+
+iOS only allows certain kind of applications to keep a connection open when in the
```
rajeevgopalakrishna, line 61: Interesting. Any pointer to how an application can qualify? Does this mean that if/when Status adds any of such features, it will qualify for this and we will then not need the proposed solution?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+Applications on iOS can also request execution time when they are in the [background](https://developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_background/updating_your_app_with_background_app_refresh)
+but it has a limited set of use cases, for example it won't schedule any time
+if the application was force quit, and generally is not responsive enough to implement
+a push notification system.
+
+Therefore Status provides a set of Push notification services that can be used
+to achieve this functionality.
+
+Because this can't be safely implemented in a privacy preserving manner, clients
+MUST be given an option not to enable push notifications and not to send push notifications.
```
rajeevgopalakrishna, line 73: ~MUST be given an option not to enable push notifications and not to send push notifications.~
How about this:
"MUST be given an option to opt-in to receiving and sending push notifications. They are disabled by default."

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+
+## Requirements
+
+The party releasing the app needs to run their own [gorush](https://github.com/appleboy/gorush)
```
rajeevgopalakrishna, line 77: ~The party releasing the app needs to run their own~
The party releasing the app, Status in this case, needs to run its own

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+
+### Gorush instance
+
+A [gorush](https://github.com/appleboy/gorush) instance MUST be publicly 
+available, this will be used only by push notification servers.
+
+### Push notification server
+
+A push notification server used by clients to register for push notification and
+sending push notifications.
```
rajeevgopalakrishna, line 90: ~A push notification server used by clients to register for push notification and sending push notifications.~
A push notification server used by clients to register for receiving and sending push notifications.

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+
+### Registering client
+
+A status client that wants to receive push notifications
```
rajeevgopalakrishna, line 94: ~status~ Status

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+
+### Sending client
+
+A status client that wants to send push notifications
```
rajeevgopalakrishna, line 98: ~status~ Status

docs/raw/push-notification-server.md
rajeevgopalakrishna: Why use `version` and not `nonce`?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+}
+```
+
+A push notification server will handle the message according to the following rules:
+
+- it MUST verify that the `signature` matches the public key of the sender.
+- it MUST verify that `token_type` is supported
+- it MUST verify that `token` is non empty
+- it MUST verify that `installation_id` is non empty
+- it MUST verify that `version` is non-zero and greater than the currently stored, if any
```
rajeevgopalakrishna, line 150: "greater than the currently stored version for the public key of the sender" I suppose?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+be sent with `error` set to `VERSION_MISMATCH`.
+
+If any other error occurs the `error` should be set to `INTERNAL_ERROR`.
+
+If the request is successful, an access token MUST be generated and sent back to the client
+with `success` set to `true`.
+
+Otherwise a response MUST be sent with `success` set to `false`.
+
+It is RECOMMENDED to use a randomly generated string for `access_token` of at least 36 characters.
```
rajeevgopalakrishna, line 171: Where do we get this requirement of `at least 36 characters` from?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+
+`request_id` should be set to the `SHA3-256` of the signature sent by the client.
+
+The request MUST be sent on the [partitioned topic][./10-waku-usage.md#partitioned-topic] of the sender.
```
rajeevgopalakrishna, line 175: ~request~ response

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+public key of the node, encrypted with this key.
+
+The content of the message MUST contain the following [protobuf record](https://developers.google.com/protocol-buffers/):
+
+```protobuf
+
+message PushNotificationOptions {
+  boolean enabled = 1;
+  repeated string allowed_user_list = 2;
+  repeated string block_user_list = 3;
```
rajeevgopalakrishna, line 116: Can we call this `blocked_user_list` to be consistent with `allowed_user_list`?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+  repeated string block_chat_list = 4;
```
rajeevgopalakrishna, line 117: Can we call this `blocked_chat_list` to be consistent with `allowed_user_list`?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+It is RECOMMENDED to use a randomly generated string for `access_token` of at least 36 characters.
+
+`request_id` should be set to the `SHA3-256` of the signature sent by the client.
+
+The request MUST be sent on the [partitioned topic][./10-waku-usage.md#partitioned-topic] of the sender.
+
+The payload of the response is:
+
+```protobuf
+message PushNotificationPreferencesResponse {
```
rajeevgopalakrishna, line 180: Would this be better named as `PushNotificationRegistrationResponse` because the response is to the registration request?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+[//]: (We can ratched and use a partitioned topic here)
+
+A client SHOULD listen for a response sent on their [partitioned topic][./10-waku-usage.md#partitioned-topic].
+
+If `success` is `true` the client has registered successfully.
+
+If `success` is `false`:
+
+- If `MALFORMED_MESSAGE` is returned, the request SHOULD not be retried without ensuring 
+  that is correctly formed.
```
rajeevgopalakrishna, line 206: retried ~without~ by ensuring that it is correctly formed.

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+- If `INTERNAL_ERROR` is returned, the request MAY be retried, but the client MUST be 
+  backoff exponentially
+- If `VERSION_MISMATCH` is returned, the client SHOULD ensure that the remote version returned in `preferences` is integrated, and the version should be incremented and the request MAY be sent again.
```
rajeevgopalakrishna, line 209: Change to "..the client SHOULD ensure that the version is incremented to beyond the remote version returned in `preferences` and then the request MAY be sent again."?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+
+A client MAY register with multiple Push Notification Servers in order to increase availability.
+
+A client SHOULD make sure that all the notification services they registered with have the same information about their devices and tokens.
+
+If no response is returned the request SHOULD be considered failed and MAY be retried with the same server or a different one, but clients MUST exponentially backoff after each trial.
+
+If the request is successful a token is returned. This SHOULD be 
+[advertised](#advertising-a-push-notification-server) as described below if no public key filtering is necessary.
```
rajeevgopalakrishna, line 218: This "public key filtering" refers to the allowed/blocked lists, correct?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+
+## Re-registering with the push notification server
+
+A client SHOULD re-register with the node if the APN or FIREBASE token changes.
```
rajeevgopalakrishna, line 222: When does this happen? I suppose such a change is triggered by the client itself?

docs/raw/push-notification-server.md
rajeevgopalakrishna: So for the PN server, this will look like a new registration request except for version check?

docs/raw/push-notification-server.md
rajeevgopalakrishna: This might also be relevant if/when we enable multiple chat keys/identities for the same user. Switching chat keys by the user should be treated the same as user logout above.

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+
+
+## Changing options
+
+This is handled in exactly the same way as re-registering above.
+
+## Unregistering from push notifications
+
+To unregister a client MUST send a `PushNotificationRegister` request as described
+above with `unregister` in `PushNotificationPreferences` set to `true`, or removing
```
rajeevgopalakrishna, line 238: Why "or" removing their device information?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+their device information.
+
+The server MUST remove all data about this user if `unregistering` is `true`,
+apart from the `hash` of the public key and the `version` of the last options,
+in order to make sure that old messages are not processed.
+
+A client MAY unregister from a server on explicit logout if multiple accounts 
+are used on a single device.
```
rajeevgopalakrishna, line 246: Why is this `MAY` and not `SHOULD`? As a user who has logged out, I would not expect notifications meant for my chat key to be delivered to this device anymore, even if there are _no_ multiple accounts used on that device. That would be a privacy leak otherwise, wouldn't it?

Won't automatically registering (based on opt-in preferences) and unregistering on user login/logout be more in line with user's expectations?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+
+MUST be sent to the server. 
+
+The server MUST be checking the signature of the message and comparing
+that with the latest `PushNotificationPreferences` for the given public key.
+
+If the requester has access to any access token, a response MUST be sent:
+
+```protobuf
+message PushNotificationQueryInfo {
```
rajeevgopalakrishna, line 308: This information should include the public key of user associated with the `access_token`, correct? Otherwise, how is a response bound to a certain query?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+## Sending a push notification
+
+When sending a push notification, only the `installation_id` for the devices targeted 
+by the message SHOULD be used.
+
+If a message is for all the user devices, all the `installation_id` known to the client MAY be used. 
+
+The number of devices MAY be capped in order to reduce resource consumption.
+
+At least 3 devices SHOULD be targeted, ordered by last activity.
```
rajeevgopalakrishna, line 339: Where does this requirement of `At least 3 devices` come from? Is it a heuristic?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+#### Data disclosed
+
+- Public key of the author
+
+### PushNotificationPreferences
+
+A push notification preferences message describes the push notification options and tokens for all the devices associated with `PublicKeyClient`.
+
+`device_tokens`: a list of `PushNotificationPreferences`, one for each device owned by the user.
+`version`: an atomically increasing number identifying the current `PushNotificationPreferences`. Any time anything is changed in the record it MUST be increased by the client, otherwise the request will not be accepted.
```
rajeevgopalakrishna, line 456: ~atomically~ monotonically

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+
+#### Data disclosed
+
+- Public key of the author
+
+### PushNotificationPreferences
+
+A push notification preferences message describes the push notification options and tokens for all the devices associated with `PublicKeyClient`.
+
+`device_tokens`: a list of `PushNotificationPreferences`, one for each device owned by the user.
```
rajeevgopalakrishna, line 455: Ah, so `device_tokens` are not really tokens but a list of `PushNotificationPreferences` which contain tokens among other things. This name was a bit confusing.

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+
+## Partitioned topic
+
+This is a modification of the [partitioned topic](./10-waku-usage.md#partitioned-topic) used by clients
+
+```golang
+var partitionsNum *big.Int = big.NewInt(5000)
+var partition *big.Int = big.NewInt(0).Mod(publicKey.X, partitionsNum)
+
+partitionTopic := "push-notifications-" + strconv.FormatInt(partition.Int64(), 10)
```
rajeevgopalakrishna, line 576: So this prefix of `push-notifications-` is the only modification?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+
+
+## Anonymous mode of operations
+
+An anonymous mode of operations MAY be provided by the client, where the 
+responsibility of propagating information about the user is left to the client,
+in order to preserve privacy.
+
+A client in anonymous mode can register with the server using a key different 
+from their identity key.
```
rajeevgopalakrishna, line 599: What is this `identity key`? Is this the chat key?

docs/raw/push-notification-server.md
```diff
@@ -0,0 +1,765 @@
...
+the access token using an ephemeral key when querying the server.
+
+The method described above effectively does not share the identity of the sender
+nor the receiver to the server, but MAY result in missing push notifications as
+the propagation of the secret is left to the client.
+
+This can be mitigated by [device syncing](./6-payloads.md), but not completely
+addressed.
+
+[//]: This section can be removed, for now leaving it here in order to help with the
```
rajeevgopalakrishna, line 620: Sorry, I don't fully understand how this works yet.

docs/raw/push-notification-server.md
rajeevgopalakrishna: Agree. Moving the terminology and high-level flow descriptions upfront followed by the detailed protocol and data structures will help. It will also help to clarify the relationship/mapping among device/client/user/keys.

rajeevgopalakrishna (commented):
Kudos on the monster spec. :)

Well written. I need to come back to this once more to understand the parts I didn't. Saving some terminology/formatting requests for the next iteration.

docs/raw/push-notification-server.md
cammellos: my bad, it's foreground service,
in terms of getting killed, that's something I have been told by devs who have looked into this more carefully https://stackoverflow.com/questions/49637967/minimal-android-foreground-service-killed-on-high-end-phone @yenda might have more info

docs/raw/push-notification-server.md
cammellos: Currently it's only for VoIP application I believe

docs/raw/push-notification-server.md
cammellos: It's a recommendation, not a requirement, but it's basically the length of an UUID, which looked safe enough as a secret

docs/raw/push-notification-server.md
cammellos: Not sure how it reads this one:
` If `MALFORMED_MESSAGE` is returned, the request SHOULD not be retried by ensuring 
  that is correctly formed.`

Did you mean also to change `SHOULD not` (maybe not should be capitalized), to `SHOULD`? In which case the meaning is slightly different, as I would like to make clear that the request SHOULD NOT be retried without modifying it, but the client might not retry it

docs/raw/push-notification-server.md
cammellos: correct

docs/raw/push-notification-server.md
cammellos: Yes, it's up  to the client to refresh the token periodically with APN/Firebase

docs/raw/push-notification-server.md
cammellos: In case you still want to keep some of your devices registered, basically you might want to unregister your android device, but not your IOS device (or you might want to unregister both, in which case the whole profile is removed).

docs/raw/push-notification-server.md
rajeevgopalakrishna: Ah right. "If MALFORMED_MESSAGE is returned, the request SHOULD NOT be retried without ensuring that it is correctly formed" makes sense.

docs/raw/push-notification-server.md
cammellos: It's what we support by default in for messaging

docs/raw/push-notification-server.md
cammellos: Yes, should we rename it?

docs/raw/push-notification-server.md
cammellos: correct

docs/raw/push-notification-server.md
cammellos: yes

docs/raw/push-notification-server.md
cammellos: The  whole thing or only the device syncing part? I can further elaborate if necessary

docs/raw/push-notification-server.md
cammellos: I think this is more of a `version` than a `nonce`, a `nonce` is literally a `number used once`.
In our case 1) we always refer to the same data-structure, which clients are supposed to pull from the server and update 2) The nonce needs to be monotonically increasing 

I guess it depends on what people are most familiar with, but this is an established pattern by distributed databases using optimistic locking and generally version is used ( elasticsearch/dynamodb)

docs/raw/push-notification-server.md
cammellos: Correct

docs/raw/push-notification-server.md
cammellos: So I think this is a bit more of a UX concern,
but there are other things to consider that are implementation dependent, and I think `MAY` is best to give clients more freedom.

The best way to handle this is not to unregister only with the server on logout, because then you get into troubles if the user logs out when offline.

Say:
1) Go offline
2) Logout
3) Someone sends you a notification
4) You go online

The notification in 3 will be dispatched on IOS, there's no way around this as we don't have that kind of local control.

Ideally when you logout, you tell IOS that you don't want to receive notifications for that token, maybe by invalidating the previous one, but each client (android/IOS and maybe desktop) will kind of have to follow their own path. Some client might not even have a logout functionality (console client).

So I think MAY is more apt as it's very much related to the specific UX of a given client, but happy to change it to a SHOULD if you feel strongly about it.

docs/raw/push-notification-server.md
cammellos: yes, good point

docs/raw/push-notification-server.md
rajeevgopalakrishna: Thanks for the clarification. Makes sense.

docs/raw/push-notification-server.md
rajeevgopalakrishna: Better to IMO.

docs/raw/push-notification-server.md
rajeevgopalakrishna: Then should we call it `chat key` here unless `identity key` is the standard terminology used elsewhere?

docs/raw/push-notification-server.md
rajeevgopalakrishna: Ok, I think I understand this now (except device syncing). The terminology (specifically, use of the words identity key, push notification identity key and secret :) threw me off earlier, sorry. 

So the difference is in the generation of a push notification key (PN key) which is different from the user's chat key (i.e. identity key). The msg sender has to be aware of this PN key to send push notifications to the recipient and the probabilistic (?) propagation of this PN key makes the receipt of PNs indeterministic. 

The PN sender may get the access token either by querying the PN server like earlier (using an ephemeral key if it wants) or directly receive it from the client by listening to contact-code topic for the PN key. The PN sender then sends a notification just like earlier by talking to the PN server which in turn talks to the gorush server.

The use of ephemeral PN-specific keys makes this mode anonymous.

cammellos: Correct,
the device syncing part is that if you have multiple devices, say A1 & A2, and you chose to propagate the token yourself (so not through the server), you need to make sure that A1 & A2 are paired, otherwise they might not be aware of each other, and therefore you might be propagating only the token for A1 to another client for example, which will only send a PN to A1.

oskarth:
Can we please do follow up discussions in separate issues? A bit difficult to follow discussion once a PR has been merged already.

cammellos:
I have opened a PR already , we can continue there https://github.com/status-im/specs/pull/143

PR #143, opened 2020-06-25
Add encrypted access tokens

cammellos:
This commit modifies the spec so that instead of passing a list of `SHA3` `public_keys` to the server for filtering, the user will pass instead a list of `access_token` encrypted with the allowed contact DH PK(bob) PK(alice).

It also removes the `blocked_user_list`, so all the filtering is done based on the `chat_id` field. 

Effectively both were used a sort of a spam prevention mechanism,  but the cost of keeping `blocked_user_list` is that a user will need to disclose their identity to the server, and is trivial to circumvent for any determined attacker, so keeping only filtering on chat_id achieves a similar naive prevention of spam without compromising the privacy of other users. 
Spam prevention is something that we don't have a solution in status app and this reflects the same behavior.

These changes effectively means that:

1) The server has no way to know which users are in `allowed_user_list`.
2) The client can query the server anonymously, so at no point they have to disclose their identity to the server.

Addressed some leftover feedback
@rajeevgopalakrishna


cammellos:
@rajeevgopalakrishna I have addressed the feedback in the closed PR here, we can probably keep the conversation here.

I have changed the protocol so that the `access_token` is generated client side and passed to the server.

I want to change how `allowed_user_list` works, and instead of passing a list of hashed public keys, the user can encrypt the token with all the public keys of the allowed users in a blob and push that to the server.

When querying users can just pull this blob and look for a token by decrypting the message, so that we don't have to communicate to the server which users are actually allowed, and conversely the users don't have to communicate their real identity to the server.
But likely I will address that in a different PR.

docs/raw/push-notification-server.md
```diff
@@ -70,32 +70,31 @@ Therefore Status provides a set of Push notification services that can be used
...
 
 Because this can't be safely implemented in a privacy preserving manner, clients
-MUST be given an option not to enable push notifications and not to send push notifications.
+MUST be given an option to opt-in to receiving and sending push notifications. They are disabled by default.
 
 ## Requirements
 
-The party releasing the app needs to run their own [gorush](https://github.com/appleboy/gorush) 
+The party releasing the app, Status in this case, needs to run its own [gorush](https://github.com/appleboy/gorush)
 publicly accessible server for sending the actual notification. This cannot be community run as it requires sharing private credentials that only status has access to.
```
0kok0: I wouldn't rule out the option that this could be run be the community at this point. I would just document that whatever entity wants to run a notification service is required to possess a certificate for the Apple Push Notification service: 



```suggestion
The party releasing the app needs to possess a certificate for the Apple Push Notification service and its has to run a [gorush](https://github.com/appleboy/gorush) publicly accessible server for sending the actual notification. 
```
Otherwise. looks good!

oskarth:
Generally seems fine to me, but I'd rather see core and security take ownership of these reviews and getting them merged. One review from core and one from security seems reasonable to me

oskarth:
Removing reviewers to avoid diffusion of responsibility, pinging @andremedeiros  @Samyoul @0kok0 @rajeevgopalakrishna for reviews

docs/raw/push-notification-server.md
```diff
@@ -70,32 +70,31 @@ Therefore Status provides a set of Push notification services that can be used
...
 
 ## Components
 
 ### Gorush instance
 
-A [gorush](https://github.com/appleboy/gorush) instance MUST be publicly 
+A [gorush](https://github.com/appleboy/gorush) instance MUST be publicly
 available, this will be used only by push notification servers.
 
 ### Push notification server
```
rajeevgopalakrishna, line 87: Can we rename this to "Push Notification _Node_" to start avoiding further usage of the term server?

docs/raw/push-notification-server.md
```diff
@@ -110,14 +109,13 @@ The content of the message MUST contain the following [protobuf record](https://
 
 ```protobuf
 
-message PushNotificationOptions {
+message PushNotificationFilterSettings {
   boolean enabled = 1;
   repeated string allowed_user_list = 2;
```
rajeevgopalakrishna: Does it make send to rename "user" to "key" here and elsewhere in this spec to distinguish the same user entity who may have multiple chat keys/identities? Given the whole protocol operates on keys, why not use that term itself? This will also be consistent with our move away from using "accounts" to "keys."

docs/raw/push-notification-server.md
```diff
@@ -202,19 +199,19 @@ If `success` is `true` the client has registered successfully.
...
+- If `VERSION_MISMATCH` is returned, the client SHOULD ensure that the version is incremented to beyond the remote version returned in preferences and then the request MAY be sent again
 
 A client MAY register with multiple Push Notification Servers in order to increase availability.
 
 A client SHOULD make sure that all the notification services they registered with have the same information about their devices and tokens.
 
 If no response is returned the request SHOULD be considered failed and MAY be retried with the same server or a different one, but clients MUST exponentially backoff after each trial.
 
-If the request is successful a token is returned. This SHOULD be 
+If the request is successful a token is returned. This SHOULD be
```
rajeevgopalakrishna: Guess this goes away now that the access token is generated client side.

docs/raw/push-notification-server.md
```diff
@@ -242,7 +239,7 @@ The server MUST remove all data about this user if `unregistering` is `true`,
 apart from the `hash` of the public key and the `version` of the last options,
 in order to make sure that old messages are not processed.
 
-A client MAY unregister from a server on explicit logout if multiple accounts 
+A client MAY unregister from a server on explicit logout if multiple accounts
```
rajeevgopalakrishna: Can we replace "accounts" with "chat keys"?

docs/raw/push-notification-server.md
```diff
@@ -297,17 +294,16 @@ message PushNotificationQuery {
 }
 ```
 
-MUST be sent to the server. 
+MUST be sent to the server.
 
-The server MUST be checking the signature of the message and comparing
-that with the latest `PushNotificationPreferences` for the given public key.
```
rajeevgopalakrishna, line 303: Replace "public key" with "chat key"?

docs/raw/push-notification-server.md
```diff
@@ -297,17 +294,16 @@ message PushNotificationQuery {
...
-
-If the requester has access to any access token, a response MUST be sent:
+A response MUST be sent:
 
 ```protobuf
 message PushNotificationQueryInfo {
   string access_token = 1;
   string installation_id = 2;
+  bytes public_key = 3;
```
rajeevgopalakrishna, line 304: Replace "public key" with "chat key"?

docs/raw/push-notification-server.md
```diff
@@ -126,13 +124,14 @@ message PushNotificationDeviceToken {
...
-  PushNotificationPreferences preferences = 4;
+  PushNotificationFilterSettings filter_settings = 4;
 }
 
 message PushNotificationPreferences {
-  repeated DeviceToken device_tokens = 1;
+  repeated PushNotificationOptions options = 1;
   uint version = 2;
   boolean unregister = 3;
+  string access_token = 4;
```
rajeevgopalakrishna, line 134: `access_token` should perhaps not be bundled under `PushNotificationPreferences`. Consider renaming or splitting message.

docs/raw/push-notification-server.md
```diff
@@ -110,14 +109,13 @@ The content of the message MUST contain the following [protobuf record](https://
...
+message PushNotificationFilterSettings {
   boolean enabled = 1;
   repeated string allowed_user_list = 2;
-  repeated string block_user_list = 3;
-  repeated string block_chat_list = 4;
+  repeated string blocked_chat_list = 3;
 }
 
-message PushNotificationDeviceToken {
+message PushNotificationOptions {
```
rajeevgopalakrishna: Sorry, I might have earlier suggested renaming "DeviceToken" here but that seems to be more appropriate than "Options". It may be a bit confusing because we have this `token` along with `access_token` now.

docs/raw/push-notification-server.md
```diff
@@ -318,21 +314,29 @@ message PushNotificationQueryResponse {
...
 
-A client MAY send request to the server using an ephemeral key at first, in order not to reveal 
-their identity.
+[//]: TODO: Add more details on the exact encryption method, AES-CTR etc
 
-If the request is not successful, they should assume that some public key filtering is
-in place at the server level and they MAY retry the request using their real identity.
+If `encrypted_access_tokens` is returned, the client SHOULD decrypt each
+token by generating a `AES` symmetric key from the Diffie–Hellman between the
+target client and the querying client.
```
rajeevgopalakrishna, line 328: ~target client and the querying client~
target client and itself.

docs/raw/push-notification-server.md
```diff
@@ -318,21 +314,29 @@ message PushNotificationQueryResponse {
...
+If a token is decrypted successfully it SHOULD be used to send push notifications.
```
rajeevgopalakrishna: (As discussed) We can add how to identify successful decryption - "If AES decryption succeeds it will return a valid uuid which is what is used for `access_token`."

docs/raw/push-notification-server.md
```diff
@@ -452,54 +470,52 @@ A `PushNotificationRegister` is used to register with a Push Notification server
...
 [//]: (Any of these can be sha3 in order not to store it on the server)
 
 `enabled`: whether the device wants to be sent push notifications
-`allowed_user_list`: a list of `SHA3-256` of public keys hex encoded as a string. If non-empty
-only those public keys will be issued a token.
-`block_user_list`: a list of `SHA3-256` of public keys hex encoded as a string. Any public key 
-in this list MUST not be issued a token.
-`block_chat_list`: a list of `SHA3-256` hashes of chat ids. Any chat id in this list will not trigger a notification.
+`allowed_user_list`: a list of `access_token` encrypted with the key generated
+by the AES key generated by Diffie–Hellman between the publisher and the allowed
```
rajeevgopalakrishna, line 504: ~encrypted with the key generated by the AES key generated~
encrypted with the AES key generated

docs/raw/push-notification-server.md
```diff
@@ -361,19 +365,18 @@ Where `message` is the encrypted payload of the message and `chat_id` is the
...
 MUST be sent.
-In such cases `ack_required` MAY bet set and the sender SHOULD be listening
-on the topic derived from the first 4 bytes of `SHA3-256(ack_required)` and with
-a waku AES symmetric encryption key of `ack_required`.
+The sender SHOULD be listening on the topic derived from the first 4 bytes of `SHA3-256(ack_key)` and with
+a waku AES symmetric encryption key of `ack_key`.
 
-If no response is received 
+If no response is received
 in 3 seconds, it MAY be retried against a different server.
```
rajeevgopalakrishna: Why is "3 seconds" a good heuristic?

docs/raw/push-notification-server.md
```diff
@@ -630,23 +643,25 @@ If no anonymous mode is used, when registering with a push notification service
...
 - The hash of the chat_ids they want to filter out
-- The hash of the public keys that the client is interested/not interested in
 
 When running in anonymous mode, the client's public key is not disclosed.
 
-If no anonymous mode is used, when querying a notification service a client discloses:
+When querying a push notification server a client will disclose:
 
-- The public key
+- The fact that is interested in sending push notification to a given client,
```
rajeevgopalakrishna, line 651: ~The fact that is~
The fact that it is

docs/raw/push-notification-server.md
```diff
@@ -720,12 +735,13 @@ Most of the request have a target, so protocol negotiation can happen. We cannot
...
-### Why ack_required?
+### Why ack_key?
 
-That's necessary to avoid duplicated push notifications.
+That's necessary to avoid duplicated push notifications and allow for the retry
+in case the notification is not successful.
 
-Deduplication of the push notification is done on the client side, to relive a bit 
-of centralization and also in order not to have to modify gorush. 
+Deduplication of the push notification is done on the client side, to relive a bit
```
rajeevgopalakrishna, line 743: ~relive~
reduce

rajeevgopalakrishna (changes requested):
More privacy now with encrypted tokens - nice work! Just a few minor (formatting) suggestions to address for now.

docs/raw/push-notification-server.md
```diff
@@ -202,19 +199,19 @@ If `success` is `true` the client has registered successfully.
 
 If `success` is `false`:
 
-- If `MALFORMED_MESSAGE` is returned, the request SHOULD not be retried without ensuring 
+- If `MALFORMED_MESSAGE` is returned, the request SHOULD NOT be retried without ensuring
   that is correctly formed.
```
rajeevgopalakrishna:
```suggestion
  that it is correctly formed.
```

docs/raw/push-notification-server.md
```diff
@@ -202,19 +199,19 @@ If `success` is `true` the client has registered successfully.
...
-- If `INTERNAL_ERROR` is returned, the request MAY be retried, but the client MUST be 
+- If `INTERNAL_ERROR` is returned, the request MAY be retried, but the client MUST be
```
rajeevgopalakrishna:
```suggestion
- If `INTERNAL_ERROR` is returned, the request MAY be retried, but the client MUST 
```

docs/raw/push-notification-server.md
rajeevgopalakrishna:
```suggestion
target client and itself.
```

docs/raw/push-notification-server.md
rajeevgopalakrishna:
```suggestion
`allowed_user_list`: a list of `access_token` encrypted with the AES key generated
 by Diffie–Hellman between the publisher and the allowed
```

docs/raw/push-notification-server.md
rajeevgopalakrishna:
```suggestion
- The fact that it is interested in sending push notification to a given client,
```

docs/raw/push-notification-server.md
rajeevgopalakrishna:
```suggestion
Deduplication of the push notification is done on the client side, to reduce a bit
```

rajeevgopalakrishna (changes requested):

docs/raw/push-notification-server.md
Samyoul: It appears that you are amending line 78 when I think you were aiming to give a suggestion for line 77

Also I'd replace "needs" with "MUST"

>The party releasing the app MUST to possess a certificate for the Apple Push Notification service and its has to run a [gorush](https://github.com/appleboy/gorush) publicly accessible server for sending the actual notification.

docs/raw/push-notification-server.md
Samyoul: I'm curious, what is the benefit of avoiding the term "server"?

docs/raw/push-notification-server.md
```diff
@@ -396,17 +399,32 @@ following data:
 Where platform is `1` for IOS and `2` for Firebase, according to the [gorush
 documentation](https://github.com/appleboy/gorush)
 
-If the `ack_required` is set, a server MUST be sending back a message:
+A server MUST be sending back a message:
```
Samyoul:
```suggestion
A server MUST return a response message:
```

docs/raw/push-notification-server.md
```diff
@@ -318,21 +314,29 @@ message PushNotificationQueryResponse {
...
+If `access_token` is returned, this SHOULD be used to send push notifications.
 
-A client MAY send request to the server using an ephemeral key at first, in order not to reveal 
-their identity.
+[//]: TODO: Add more details on the exact encryption method, AES-CTR etc
 
-If the request is not successful, they should assume that some public key filtering is
-in place at the server level and they MAY retry the request using their real identity.
+If `encrypted_access_tokens` is returned, the client SHOULD decrypt each
+token by generating a `AES` symmetric key from the Diffie–Hellman between the
```
Samyoul:
```suggestion
If `encrypted_access_tokens` are returned, the client SHOULD decrypt each
token by generating an `AES` symmetric key from the Diffie–Hellman between the
```

docs/raw/push-notification-server.md
```diff
@@ -318,21 +314,29 @@ message PushNotificationQueryResponse {
...
 Otherwise a response MUST NOT be sent.
 
-Multiple queries MAY be sent for the same token from different public keys to
-increase deniability.
+If `allowed_user_list` is not set `access_token` MUST be set and `encrypted_access_tokens` MUST NOT
+be set.
+
+If `allowed_user_list` is set `encrypted_access_tokens` MUST be set and `access_token` MUST NOT be set.
+
+If `access_token` is returned, this SHOULD be used to send push notifications.
```
Samyoul: To avoid an ambiguous "this".

```suggestion
If `access_token` is returned, the `access_token` SHOULD be used to send push notifications.
```

docs/raw/push-notification-server.md
Samyoul: Does this sentence describe communication between a Push Notification Server and a client node? Or are we talking about p2p comms between clients?

The section seems to be about server / client querying, but this sentence seems to be about p2p comms. Perhaps add a sentence or subsection header to describe the difference between directly querying the server and client to client querying. Though I may be misunderstanding / missing the part where this distinction is made

docs/raw/push-notification-server.md
```diff
@@ -452,54 +470,52 @@ A `PushNotificationRegister` is used to register with a Push Notification server
 
 A push notification preferences message describes the push notification options and tokens for all the devices associated with `PublicKeyClient`.
 
-`device_tokens`: a list of `PushNotificationPreferences`, one for each device owned by the user.
-`version`: an atomically increasing number identifying the current `PushNotificationPreferences`. Any time anything is changed in the record it MUST be increased by the client, otherwise the request will not be accepted.
+`options`: a list of `PushNotificationOptions`, one for each device owned by the user.
+`version`: an monotonically increasing number identifying the current `PushNotificationPreferences`. Any time anything is changed in the record it MUST be increased by the client, otherwise the request will not be accepted.
```
Samyoul: I has to Google that "monotonically" term.

docs/raw/push-notification-server.md
```diff
@@ -630,23 +643,25 @@ If no anonymous mode is used, when registering with a push notification service
...
-- The hash of the public keys that the client is interested/not interested in
 
 When running in anonymous mode, the client's public key is not disclosed.
 
-If no anonymous mode is used, when querying a notification service a client discloses:
+When querying a push notification server a client will disclose:
 
-- The public key
+- The fact that is interested in sending push notification to a given client,
+  but their public key is not disclosed
```
Samyoul:
```suggestion
- That it is interested in sending push notification to another given client,
  but the querying client's public key is not disclosed
```

Samyoul (approved):
Great work. Just a few points of clarification that I feel need addressing.

Also sorry for the late review @cammellos , I actually read through this whole PR on Monday, but got sidetracked with Nim 😬

docs/raw/push-notification-server.md
cammellos: I don't mind either, technically the lines are a bit blurred, as they both operate as waku clients, but there's definitely a client/server relationship, so happy to go with any that we decide

docs/raw/push-notification-server.md
cammellos: I might do that in a separate PR if it's ok, I will be getting rid of a few messages in order to simplify and we can shuffle things around at that point if that's ok.

docs/raw/push-notification-server.md
cammellos: This describe a client querying a server,
although the definition of a server is a bit lose, any client can act as a push notification node(server) if it  wishes to do so, but that's probably a different matter.

Basically you pull the encrypted data from the server, and then you decrypt it taking into consideration your pk and the pk of the user who has registered with the server, but no direct communication between the two clients.

docs/raw/push-notification-server.md
cammellos: Sounds like a good amount of time :)
I was thinking 1 sec for sending, 1 sec for receiving + 1 sec for processing time should give enough room in most cases, but can be changed

docs/raw/push-notification-server.md
cammellos: That's a bit tricky,
I'd say, let's park the renaming for now and we will do in a different PR if it that's  ok with you.
The thing I am not sure is that not always this will be chat keys, for example in anonymous mode this are just keys that a user creates specifically for push notifications, but they are not chat keys.

docs/raw/push-notification-server.md
rajeevgopalakrishna: Ok, just that it's confusing/misleading to see a client/server relationship in a p2p network (e.g. mailserver vs history node). We can cast this as a special type of node that is queried by others today with the potential to merge this functionality into other nodes in network later.

docs/raw/push-notification-server.md
rajeevgopalakrishna: Good rule of thumb :) but perhaps technically defendable if we parameterise this in the spec so it can be changed depending on network congestion or node spec.

docs/raw/push-notification-server.md
cammellos: ok, so I have updated the references where we refer 100% to a chat key

docs/raw/push-notification-server.md
rajeevgopalakrishna: We could think of these as capabilities (history, relaying, notifications etc.) that can be implemented/enabled by various nodes without enforcing a strict client-server relationship.

docs/raw/push-notification-server.md
cammellos: how could we phrase this?
`in a reasonable amount of time given the current network conditions`?

rajeevgopalakrishna: Ideally have a config parameter `retry_period = 3s` with exponential backoff?

cammellos: We can make it configurable in the implementation, that's fine, what if instead we word it as:

```
If no response is received
a client SHOULD wait at least 3 seconds, after which the request MAY be retried against a different server
```
That way we only cap for a reasonable minimum and the client is free to wait longer if they wish to do so, taking into accounts network condidtions etc?

rajeevgopalakrishna: Sounds good, thanks.

cammellos:
@rajeevgopalakrishna @Samyoul I should have addressed all the outstanding feedback.
There's still some considerable changes that I need to make, but I will do in a different PR (getting rid of some of the messages, specify encryption etc), so there will be still time for feedback at a later round.
thanks

rajeevgopalakrishna (approved):

PR #145, opened 2020-07-10
Added EmojiReaction payload types

Samyoul:
Added emoji reaction payload specifications.

For reference see:

- [`status-go` implementation](https://github.com/status-im/status-go/pull/1999)
- [`status-react` implementation](https://github.com/status-im/status-react/pull/10912)
- [UI process flows](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=1239%3A0)

I've added these specs to the draft version of `6-payloads` because I saw images in drafts and presumed that emoji reactions would also be considered on the same level or lower of spec stability.

### DIFF
--- a/docs/draft/6-payloads.md
+++ b/docs/draft/6-payloads.md
@@ -35,6 +35,8 @@ as various clients created using different technologies.
      - [Content types](#content-types)
        - [Sticker content type](#sticker-content-type)
        - [Image content type](#image-content-type)
+       - [EmojiReaction content type](#emojireaction-content-type)
+       - [EmojiReactionRetraction content type](#emojireactionretraction-content-type)
      - [Message types](#message-types)
      - [Clock vs Timestamp and message ordering](#clock-vs-timestamp-and-message-ordering)
      - [Chats](#chats)
@@ -109,6 +111,8 @@ message ChatMessage {
   oneof payload {
     StickerMessage sticker = 9;
     ImageMessage image = 10;
+    EmojiReaction emoji_reaction = 11;
+    EmojiReactionRetraction emoji_reaction_retraction = 12;
   }
 
   enum MessageType {
@@ -128,6 +132,8 @@ message ChatMessage {
     // Only local
     SYSTEM_MESSAGE_CONTENT_PRIVATE_GROUP = 6;
     IMAGE = 7;
+    EMOJI_REACTION = 8;
+    EMOJI_REACTION_RETRACTION = 9;
   }
 }
 ```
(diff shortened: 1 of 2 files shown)


docs/draft/6-payloads.md
```diff
@@ -144,7 +150,7 @@ message ChatMessage {
 | 6 | chat_id | `string` | The local ID of the chat the message is sent to |
 | 7 | message_type | `MessageType` | The type of message, different for one-to-one, public or group chats |
 | 8 | content_type | `ContentType` | The type of the content of the message | 
-| 9 | payload | `Sticker|Image|nil` | The payload of the message based on the content type |
+| 9 | payload | `Sticker` I `Image` I `EmojiReaction` I `EmojiReactionRetraction` I `nil` | The payload of the message based on the content type |
```
Samyoul: `|` would be parsed by the markdown renderer as a new table cell token, even though it was inside backticks. I've replaced `|` with `I` because markdown font renders `I` as "I"

oskarth:
Generally looks good to me, I'll leave this review for Status and core though :) @status-im/security-internal  might want to sanity too

docs/draft/6-payloads.md
```diff
@@ -191,6 +199,49 @@ message ImageMessage {
...
+
+This specification RECOMMENDS that the UI/UX implementation of sending `EmojiReactionRetraction`s requires only a single
+click operation, as users have an expectation that emoji reaction removals are effortless and simple to perform.  
+
+A `ChatMessage` with `EMOJI_REACTION_RETRACTION` `Content/Type` MUST also specify the `EmojiReactionRetraction` field
+`emoji_reaction_id`
+
+```protobuf
+message EmojiReactionRetraction {
+  string emoji_reaction_id = 1;
```
rajeevgopalakrishna: Does `emoji_reaction_id` refer to the corresponding `message_id` from `message EmojiReaction`?

rajeevgopalakrishna (commented):
Looks good to me. Curious - can we apply the same retraction concept from here to messages as well?

oskarth:
Meta comment

> I've added these specs to the draft version of 6-payloads because I saw images in drafts and presumed that emoji reactions would also be considered on the same level or lower of spec stability.

Once it is live it should be marked as stable, because it is a contract with end users as well. I believe @cammellos has done some draft/stable splitting of specs in past to make this distinction clear, so that might be a a good place to look re that.

docs/draft/6-payloads.md
Samyoul: Yes.

So, `message EmojiReaction` will generate its own id and the `EmojiReactionRetraction` targets the id of the `EmojiReaction`. This approach will allow messages to have multiple reactions.

Samyoul:
@rajeevgopalakrishna 

>Looks good to me. Curious - can we apply the same retraction concept from here to messages as well?

We certainly could. In my mind this is the only way to "remove" messages from a p2p system. We need to send some kind of kill signal to peers that would honour the retraction of a message.

Samyoul:
@oskarth 

> Once it is live it should be marked as stable, because it is a contract with end users as well. I believe @cammellos has done some draft/stable splitting of specs in past to make this distinction clear, so that might be a a good place to look re that.

Thank you for that meta comment, this is useful information to know.

hesterbruikman:
Any changes based on the current status-go implementation @Samyoul ?

rasom (approved):
nice spec!

PR #146, opened 2020-07-23
Add grant, separate devices and ephemeral keys usage

cammellos:
Adds the latest changes:

- Add a grant field that Bob can use to validate that the server is authorized to send Alice push notifications
- Use of ephemeral keys 
- Use metadata wrapper from status


rasom (approved):
implementation is already merged, right?

PR #148, opened 2020-09-03
Add mentions push notifications

cammellos:
Adds support for push notification and mentions in protobuf.

### DIFF
--- a/docs/raw/push-notification-server.md
+++ b/docs/raw/push-notification-server.md
@@ -386,6 +386,13 @@ message PushNotification {
   bytes public_key = 3;
   string installation_id = 4;
   bytes message = 5;
+  PushNotificationType type = 6;
+  enum PushNotificationType {
+    UNKNOWN_PUSH_NOTIFICATION_TYPE = 0;
+    MESSAGE = 1;
+    MENTION = 2;
+  }
+  bytes author = 7;
 }
 
 message PushNotificationRequest {
@@ -399,6 +406,7 @@ A `PushNotificationRequest` message MUST be wrapped in a [`ApplicationMetadataMe
 Where `message` is the encrypted payload of the message and `chat_id` is the
 `SHAKE-256` of the `chat_id`.
 `message_id` is the id of the message
+`author` is the `SHAKE-256` of the public key of the sender.
 
 If multiple server are available for a given push notification, only one notification
 MUST be sent.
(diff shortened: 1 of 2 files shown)


oskarth (approved):
Looks fine to me. Perhaps want to update changelog depending on what state this spec is in. Is this spec still in raw? Or is it in a draft/stable state?

docs/raw/push-notification-server.md
```diff
@@ -514,6 +518,9 @@ Any chat id in this list will not trigger a notification.
 `unregister`: whether the account should be unregistered
 `grant`: the grant for this specific server
 `allow_from_contacts_only`: whether the client only wants push notifications from contacts
+`apn_topic`: the APN topic for the push notification
+`block_mentions`: whether the client does not want to be notified on mentions
```
rajeevgopalakrishna, line 530: Nit: does it make sense to flip this to `allow_mentions` to be consistent with `allow_from_contacts_only`?

docs/raw/push-notification-server.md
```diff
@@ -514,6 +518,9 @@ Any chat id in this list will not trigger a notification.
...
+`allowed_mentions_chat_list`:  a list of `SHA2-256` hashes of chat ids where we want to receive mentions
```
rajeevgopalakrishna, line 531: Nit: does it make sense to flip this to `blocked_mentions_chat_list` to be consistent with `blocked_chat_list`? Or is this preferred because the default assumption is user expects mention notifications turned off in most chats unless authorized?

docs/raw/push-notification-server.md
rajeevgopalakrishna, line 530: How does `block_mentions` interact with `allow_from_contacts_only`?

rajeevgopalakrishna (approved):

docs/raw/push-notification-server.md
cammellos: it's to ease off migration, if we use `block_mentions` (defaults to false in the client) we don't have to migrate the records on the server

docs/raw/push-notification-server.md
cammellos: that's correct, essentially only if you joined a chat you want to receive a mention, so we need a list of joined chats

docs/raw/push-notification-server.md
cammellos: if `allow_from_contacts_only` is only, only mentions from contacts will be shown

Samyoul (approved):
Wow, the review of the actual implementation of this spec was significantly more work. 😄

decanus:
@cammellos what is the state of this?

cammellos:
@decanus merged, fell of my radar, thank you

PR #149, opened 2020-09-08
Updated draft payload spec to stable

Samyoul:
All of the differences in the current draft are now included in the shipped version of the status app.

### DIFF
--- a/docs/stable/6-payloads.md
+++ b/docs/stable/6-payloads.md
@@ -6,11 +6,11 @@ title: 6/PAYLOADS
 
 # 6/PAYLOADS
 
-> Version: 0.3
+> Version: 0.6
 >
 > Status: Stable
 >
-> Authors: Adam Babik <adam@status.im>, Andrea Maria Piana <andreap@status.im>, Oskar Thorén <oskar@status.im> (alphabetical order)
+> Authors: Adam Babik <adam@status.im>, Andrea Maria Piana <andreap@status.im>, Oskar Thorén <oskar@status.im>, Samuel Hawksby-Robinson <samuel@status.im> (alphabetical order)
 
 ## Abstract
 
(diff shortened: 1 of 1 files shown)


docs/stable/6-payloads.md
```diff
@@ -62,45 +65,16 @@ The node wraps all payloads in a [protobuf record](https://developers.google.com
 record:
 
 ```protobuf
-message ApplicationMetadataMessage {
-  bytes signature = 1;
-  bytes payload = 2;
-
-  Type type = 3;
-
-  enum Type {
```
oskarth, line 71: What happened to all of these?

Samyoul: Hmm, that is a very good question, I don't know. They are present in `stable/6-payload`, I'll check the commits on `draft/6-payload` and see what happened.

cammellos: Generally I think the workflow raw -> draft-> stable is a bit cumbersome, if multiple people are editing the same files, as those edits will be separate and merging changes is difficult.
We either find a suitable git workflow or simplify, at this stage imo we can get away with a single spec file and mark the relevant sections as raw/draft/stable, to avoid these kind of issues, but don't feel strongly about it.

Samyoul: I agree with you, it would be a much simpler process to have a single spec file and have dedicated raw/draft/stable sections.

How do you imagine the approach:

**Approach A**

```
# Stable

## A Stable Section

## Another Stable Section

## Yet another Stable Section

---

# Draft

## A Draft Section

## Another Draft Section

## Yet another Draft Section

---

# Raw

## A Raw Section

## Another Raw Section

## Yet another Raw Section

```

or 

**Approach B**

```
## A Section About A Thing [stable]

## A Section About Another Thing [draft]

## A Section About Yet Another Thing [stable]

## A Section Discussing Something [stable]

## A Section On The Topic of A Thing [raw]
```

Benefits of **Approach A**, all sections of the same stability are together and not mixed.

Benefits of **Approach B** less git diffs caused by moving text within a document. Only the spec state tag is changed.

cammellos: I'd favor maybe Approach B, but we can try either and see which one is more comfortable, or keep things as they are, I don't have a strong opinion.

oskarth: FWIW in Vac specs we have something similar to approach B.

There is a general state of the spec as a whole (raw/draft/stable), then specific (optional) sections might be marked as raw/draft etc.

This makes it clear what the state of the spec as a whole is, and points to behaviours that are still experimental.

oskarth:
What's the state of this PR?

PR #150, opened 2020-09-09
Introduction of ChatIdentity

Samyoul:
## What has changed?

I've introduced the payload type of `ChatMessageIdentity`. This payload type will be used to optionally broadcast user determined identity data about the user.

The immediate use case of the `ChatMessageIdentity` is user determined profile images and display names.

Additionally the change marks `ChatMessage.ens_name` field as deprecated, `ens_name` will become part of the `ChatMessageIdentity`.

## Why make the change?

User's want to be able to have user determined identity. User's want the option to overwrite the default "3 word name" and identicon with a representation that they have control over.

## Note

The use of `ChatMessageIdentity` is 100% optional and users may choose not use any part of the `ChatMessageIdentity` and therefore default to the "3 word name" and identicon.

Also see:

- [Profile image `epic`](https://github.com/status-im/status-react/issues/11047)
- [Avatar discuss post](https://discuss.status.im/t/profile-avatars-an-exploration/1948)

### DIFF
--- a/docs/draft/6-payloads.md
+++ b/docs/draft/6-payloads.md
@@ -261,7 +261,7 @@ message is supposed to be displayed last in a chat. This is where the basic algo
 as it's only meant to order causally related events.
 
 The status client therefore makes a "bid", speculating that it will beat the current chat-timestamp, s.t. the status client's
-Lamport timestamp format is: `clock = `max({timestamp}, chat_clock + 1)`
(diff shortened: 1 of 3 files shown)


Samyoul:
Note, this PR assumes the merging of #149

oskarth:
What's the state of this PR?

PR #154, opened 2021-06-21
Add spec for the contact request refactor

jrainville:
Specs Contacts, the contact requests and the way to persist them on the network.

### DIFF
--- a/docs/draft/6-payloads.md
+++ b/docs/draft/6-payloads.md
@@ -39,17 +39,24 @@ as various clients created using different technologies.
        - [Audio content type](#audio-content-type)
      - [Message types](#message-types)
      - [Clock vs Timestamp and message ordering](#clock-vs-timestamp-and-message-ordering)
+     - [Signatrue](#signature)
      - [Chats](#chats)
    - [Contact Update](#contact-update)
      - [Payload](#payload-2)
      - [Contact update](#contact-update-1)
+   - [Contact](#contact)
+     - [Payload](#payload-3)
+     - [Contact Request](#contact-request)
+       - [Backwards compatibility](#backwards-compatibility)
+     - [Contacts persistence](#contacts-persistence)
+       - [Contacts Sync](#contacts-sync)
    - [EmojiReaction](#emojireaction)
    - [SyncInstallationContact](#syncinstallationcontact)
-     - [Payload](#payload-3)
-   - [SyncInstallationPublicChat](#syncinstallationpublicchat)
      - [Payload](#payload-4)
-   - [PairInstallation](#pairinstallation)
+   - [SyncInstallationPublicChat](#syncinstallationpublicchat)
      - [Payload](#payload-5)
+   - [PairInstallation](#pairinstallation)
+     - [Payload](#payload-6)
    - [MembershipUpdateMessage and MembershipUpdateEvent](#membershipupdatemessage-and-membershipupdateevent)
  - [Upgradability](#upgradability)
  - [Security Considerations](#security-considerations)
@@ -119,6 +126,11 @@ message ChatMessage {
     AudioMessage audio = 11;
   }
 
+  // Signature of the receiving user. Only sent in one on one messages.
+  // That signature is received when teh other user sends a contact request.
+  // It is used to prove that we were once contacts
+  string signature = 12;
+
   enum ContentType {
     UNKNOWN_CONTENT_TYPE = 0;
     TEXT_PLAIN = 1;
@@ -147,6 +159,7 @@ message ChatMessage {
 | 7 | message_type | `MessageType` | The type of message, different for one-to-one, public or group chats |
 | 8 | content_type | `ContentType` | The type of the content of the message | 
 | 9 | payload | `Sticker` I `Image` I `Audio` I `nil` | The payload of the message based on the content type |
+| 10 | signature | `string` | Signature of the receiving user |
 
 #### Content types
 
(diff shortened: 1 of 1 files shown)


docs/draft/6-payloads.md
```diff
@@ -275,6 +288,16 @@ Messages with a `clock` less than `120` seconds under the Whisper/Waku timestamp
 
 The node uses `clock` value for the message ordering. The algorithm used, and the distributed nature of the system produces casual ordering, which might produce counter-intuitive results in some edge cases. For example, when a user joins a public chat and sends a message before receiving the exist messages, their message `clock` value might be lower and the message will end up in the past when the historical messages are fetched.
 
+#### Signature
```
cammellos: I think we need to define what the signature is composed of.

Basically we need to prove that a user really wanted to add another user to contacts.
So something like Sign(Keccak256(PK1 + PK2 + some custom bytes to make sure the signature can't be reused for a differet purpose)

Also probably we should name the field something a bit more meaningful

docs/draft/6-payloads.md
```diff
@@ -319,6 +342,80 @@ A client SHOULD send a `ContactUpdate` to all the contacts each time:
...
+A `Contact` does not mean a "friend". Any account from which we receive a `ContactUpdate` is added to the table.
+
+Users are considered "Friends" when they are **mutual** contacts, meaning that both of them have the `added` and `requestReceived` boolean set as `true`, which then triggers setting `mutualContact` to `true`. This is done through `ContactRequest`s.
+
+#### Payload
+
+| Field | Name | Type | Description |
+| ----- | ---- | ---- | ---- |
+| 1 | id | `string` |  ID of the contact. Hex-encoded public key (prefixed with 0x). |
+| 2 | address | `string` | Ethereum address of the contact |
```
cammellos: I think this is a legacy field, we used to have it once wallet was derived by the public key, but in case pk is enough.

docs/draft/6-payloads.md
```diff
@@ -319,6 +342,80 @@ A client SHOULD send a `ContactUpdate` to all the contacts each time:
...
+| 3 | name | `string` | Contact's ENS name |
+| 4 | ensVerified | `bool` | Whether the name of the contact is verified |
+| 5 | alias | `string` | Contact's generated username |
+| 6 | identicon | `string` | Identicon generated from public key |
```
cammellos: No need to send this over, as it can be generated by the pk

docs/draft/6-payloads.md
```diff
@@ -319,6 +342,80 @@ A client SHOULD send a `ContactUpdate` to all the contacts each time:
...
+| 7 | lastUpdated | `uint64` | Last time an update was received from the contact |
+| 8 | blocked | `bool` | Whether the contact is blocked |
+| 9 | added | `bool` | Whether the contact is added |
+| 10 | requestReceived | `bool` | Whether the contact requested us to add them |
+| 11 | mutualContact | `bool` | Whether the contact is mutual |
+| 12 | deviceInfo | `[]ContactDeviceInfo` |  |
```
cammellos: This might also be legacy and not used anymore

docs/draft/6-payloads.md
```diff
@@ -319,6 +342,80 @@ A client SHOULD send a `ContactUpdate` to all the contacts each time:
...
+
+#### Payload
+
+| Field | Name | Type | Description |
+| ----- | ---- | ---- | ---- |
+| 1 | id | `string` |  ID of the contact. Hex-encoded public key (prefixed with 0x). |
+| 2 | address | `string` | Ethereum address of the contact |
+| 3 | name | `string` | Contact's ENS name |
+| 4 | ensVerified | `bool` | Whether the name of the contact is verified |
+| 5 | alias | `string` | Contact's generated username |
```
cammellos: No need to send this over as generated by the pk

docs/draft/6-payloads.md
```diff
@@ -319,6 +342,80 @@ A client SHOULD send a `ContactUpdate` to all the contacts each time:
 
 A client SHOULD also periodically send a `ContactUpdate` to all the contacts, the interval is up to the client, the Status official client sends these updates every 48 hours.
 
+### Contact
+
+`Contact` is a representation of other accounts encountered on the app. They are added through `ContactUpdate`s propagated.
+A `Contact` does not mean a "friend". Any account from which we receive a `ContactUpdate` is added to the table.
+
+Users are considered "Friends" when they are **mutual** contacts, meaning that both of them have the `added` and `requestReceived` boolean set as `true`, which then triggers setting `mutualContact` to `true`. This is done through `ContactRequest`s.
```
cammellos: When is this sent exactly? 
the payload below looks ok if it's to be sent to your devices for syncing contacts, but otherwise is too many info for sending to other users.

cammellos (commented):
Overall looks good, nice work and thanks for writing this up!
I have some questions/comments about the `ContactUpdate` section.

The spec is a bit implementation heavy (we probably don't want to talk about tables etc, as clients might use different databases for example), but rather just behaviors.

jakubgs:
Is this PR dead?

jrainville:
@cammellos sorry for the late reply. I was on vacation when you did the changes so I missed it and before that, this feature was not a priority so I had pushed it back.

Anyway, I rebased the PR, fixed a typo and also reviewed your changes.
It looks good, my only question would be what happens with the "contact persistence" part? I saw that you removed it. Is it because it was too implementation heavy or just because it is no longer wanted?

I can understand if it is the latter, since it would have been heavy on the network and also harder to implement. The contact signature is easier and also does solve the 1-1 chat issues we have.
However, we will still have the issue of losing all your contacts on account import. Is it fixed using the normal sync instead (between devices)?

cammellos:
@jrainville 
thanks for the review,
> It looks good, my only question would be what happens with the "contact persistence" part? I saw that you removed it. Is it because it was too implementation heavy or just because it is no longer wanted?

We have now implemented contact backup, it hasn't been specced yet, but it's already in develop (and likely in desktop), and it works slightly differently, so I thought I'd remove it

jrainville:
Turns out I can't approve my own PR 😅 
Anyway, merge whenever you want.

jrainville:
Very stale

Issue #155, opened 2021-08-03
Improve spec flow process


oskarth:
*This issue is basically about cleaning up the process for Status specs similarly to what has been done for Vac specs**.

# Problem

Currently we use the COSS model (https://github.com/unprotocols/rfc/blob/master/2/README.md) for change management.

However, we don't do so fully and have some ad hoc processes. This complicates things needlessly, around versioning, breaking urls, etc.

## Proposed solution

Implement it more fully and document that this is the case.

## Notes

Specific things that stand out to me:

1. Lack of editor/contributor distinction
2. A bit too much ceremony for draft changes
3. Random non cosmetic changes to stable
4. Lack of canonical URLs leading to breaking urls
5. Not retiring unused specs
6. General ad hoc flow documented in a bunch of different places

Also see https://github.com/vacp2p/specs/issues/234

See how it works for Vac RFCs here https://rfc.vac.dev/ and the required changes https://github.com/vacp2p/rfc/issues/235
See https://github.com/vacp2p/rfc/issues/235

PR #160, opened 2021-11-15
ci: add Jenkinsfile and deploy script

jakubgs:
Necessary to build in our own CI and publish via Caddy-Git.


Jenkinsfile
```diff
@@ -0,0 +1,47 @@
...
+      numToKeepStr: '20',
+      daysToKeepStr: '30',
+    ))
+  }
+
+  environment {
+    GIT_AUTHOR_NAME = 'status-im-auto'
+    GIT_AUTHOR_EMAIL = 'auto@status.im'
+    /* Dev site deployment. */
+    DEV_SITE = 'dev.vac.dev'
```
oskarth: Huh? Status specs should be separate from Vac.dev

oskarth: Typo maybe?

jakubgs: Oh, that's just a leftover because I copied the `Jenkinsfile`. There is no `dev` in this case since this repo doesn't have a `develop` branch, I'll remove that.

oskarth (approved):

jakubgs:
I also split the docs to use three different `permalink` paths:
- `/raw/`
- `/draft/`
- `/stable/`

Otherwise docs like these clash and only one is accessible:
```
docs/draft/3-whisper-usage.md
docs/stable/3-whisper-usage.md
```
```
docs/draft/6-payloads.md
docs/stable/6-payloads.md
```

jakubgs:
This PR ties into this issue which references broken links: https://github.com/status-im/specs/issues/155

Samyoul (approved):

jakubgs:
I thought `/stable` was better because it's more clear about state, and `/spec` is redundant because the domain is already `specs.status.im`, but as Oskar points out that it changes URLs that might be embedded somewhere I've switched it back to `/spec` and renamed the folder from `stable` to `spec`.

jakubgs:
Also, moved `15/NOTIFICATIONS` to `draft` because it was never made stable.

PR #162, opened 2021-12-13
Draft: Specification for the Community History Problem (MVP)

0x-r4bbit:

PR #164, opened 2022-01-17
Raw/serving community history mvp v2

0x-r4bbit:
This the same as #162 but without all of the comments (#162 is getting too big to load).

### DIFF
--- a/docs/raw/serving-community-history.md
+++ b/docs/raw/serving-community-history.md
@@ -34,9 +34,14 @@ This specification describes how community owners archive historical message dat
 - [Exporting messages for bundling](#exporting-messages-for-bundling)
 - [Message history archives](#message-history-archives)
   - [WakuMessageHistoryArchive](#wakumessagehistoryarchive)
+- [Message history archive index](#message-history-archive-index)
+  - [WakuMessaegArchiveIndex](#wakumessagearchiveindex)
+- [Creating message archive torrents](#creating-message-archive-torrents)
+  - [Ensuring reproducible data pieces](#ensuring-reproducible-data-pieces)
+    - [Example: without padding](#example-without-padding)
+    - [Example: with padding](#example-with-padding)
+- [Seeding message history archives](#seeding-message-history-archives)
 - [Creating magnet links](#creating-magnet-links)
-- [Bundling history archives into archive indices](#bundling-history-archives-into-archive-indices)
-  - [WakuMessageArchiveMetadata](#wakumessagearchivemetadata)
 - [Message archive distribution](#message-archive-distribution)
 - [Canonical message histories](#canonical-message-histories)
 - [Fetching messsage archives](#fetching-message-archives)
@@ -70,7 +75,8 @@ The following terminology is used throughout this specification:
 
 ## Supported platforms
 
-Creating, distributing, downloading and unpacking community history archives SHOULD only be supported in Status desktop clients. Status mobile clients SHOULD NOT implement this functionality due to bandwidth and storage limitations.
+Creating, distributing, downloading and unpacking community history archives SHOULD only be supported in Status desktop clients. Status mobile are out of scope for the MVP of this service. However we may consider bringing this service to Status Mobile in the future.
+
 
 ## Requirements / Assumptions
 
@@ -100,8 +106,8 @@ The following is a high-level overview of the user flow and features this specif
 Community owner nodes go through the following (high level) process to provide community members with message histories (assumes community owner node is available 24/7):
 
 1. Community owner creates a Status community (previously known as [org channels](https://github.com/status-im/specs/pull/151))
-2. Community owner enables community history archive support (possibly on by default but can be turned off as well - see [UI feature spec](https://github.com/status-im/feature-specs/pull/36))
-3. A special type of channel to exchange metadata about the archival data is created
+2. Community owner doesn't disable community history archive support (on by default but can be turned off as well - see [UI feature spec](https://github.com/status-im/feature-specs/pull/36))
+3. A special type of channel to exchange metadata about the archival data is created, this channel should not be visible in the user interface
 4. Community owner invites members and creates additional channels
 5. Community owner node receives messages and stores them into local database
 6. After 7 days, the community owner node exports and compresses last 7 days worth of messages from database and bundles it together with a [message archive index](#waku-message-archive-index) into a torrent, from which it then creates a magnet link ([Magnet URI scheme](https://en.wikipedia.org/wiki/Magnet_URI_scheme), [Extensions for Peers to Send Metadata Files](https://www.bittorrent.org/beps/bep_0009.html)) 
@@ -224,8 +230,6 @@ A torrent's source folder MUST contain the following two files:
 - `index` - Contains the protobuf encoded message history archive index
 
 Community owner nodes SHOULD store these files in a dedicated folder that is identifiable via the community id.
-
-Community owner nodes MAY remove older torrent files that were generated for previous message history archives, after a new torrent was created.
 
 ### Ensuring reproducible data pieces
 
--- a/docs/raw/serving-community-history.md
+++ b/docs/raw/serving-community-history.md
@@ -1,0 +1,315 @@
+---
+permalink: /raw/17
+title: 17/SERVING-COMMUNITY-HISTORY-MVP
+parent: Raw specs
+layout: default
+---
+
+# 17/SERVING-COMMUNITY-HISTORY-MVP
+
+> Version: 0.1
+>
+> Status: Raw
+>
+> Authors: Pascal Precht <pascal@status.im>
+
+## Abstract
+
+Messages are stored permanently by store nodes ([11/WAKU-MAILSERVER](/spec/11), or [13/WAKU2-STORE](https://rfc.vac.dev/spec/13/)) for up to 30 days. Messages older than that are no longer provided by store nodes, making it impossible for other nodes to request historical messages older than that. This is especially problematic in the case of Status communities, where recently joined members of a community aren't able to request complete message histories of the community channels.
+
+This specification describes how community owners archive historical message data of their communities beyond the time range limit provided by store nodes using the [BitTorrent](https://bittorrent.org) protocol. It also describes how the archives are distributed to community members via the Status network, so they can fetch them and get access to a complete message history.
+
+## Table of Contents
+
+- [Abstract](#abstract)
+- [Table of Contents](#table-of-contents)
+- [Terminology](#terminology)
+- [Supported platforms](#supported-platforms)
+- [Requirements / Assumptions](#requirements--assumptions)
+- [Overview](#overview)
+  - [Serving community history archives](#serving-community-history-archives)
+  - [Serving archives for missed messages](#serving-archives-for-missed-messages)
+  - [Receiving community history archives](#receiving-community-history-archives)
+- [Storing live messages](#storing-live-messages)
+- [Exporting messages for bundling](#exporting-messages-for-bundling)
+- [Message history archives](#message-history-archives)
+  - [WakuMessageHistoryArchive](#wakumessagehistoryarchive)
+- [Creating magnet links](#creating-magnet-links)
+- [Bundling history archives into archive indices](#bundling-history-archives-into-archive-indices)
+  - [WakuMessageArchiveMetadata](#wakumessagearchivemetadata)
+- [Message archive distribution](#message-archive-distribution)
+- [Canonical message histories](#canonical-message-histories)
+- [Fetching messsage archives](#fetching-message-archives)
+  - [Downloading message archive indices](#downloading-message-archive-indices)
+  - [Downloading individual archives](#downloadinng-individual-archives)
+- [Storing historical messages](#storing-historical-messages)
+- [Considerations](#considerations)
+  - [Bandwidth consumption](#bandwidth-consumption)
+  - [Multiple community owners](#multiple-community-owners)
+
+## Terminology
+
+The following terminology is used throughout this specification:
+
+| Name                 | References |
+| -------------------- | --- |
+| Waku node            | An Ethereum node with Waku V1 enabled, or a [10/WAKU2](https://rfc.vac.dev/spec/10/) node that implements [11/WAKU2-RELAY](https://rfc.vac.dev/spec/11/)|
+| Store node           | A Waku node that implements [11/WAKU-MAILSERVER](/spec/11) or [13/WAKU2-STORE](https://rfc.vac.dev/spec/13/) respectively |
+| Waku network         | A group of Waku nodes connected through the internet connection and forming a graph |
+| Community owner      | A Status user that owns a Status community |
+| Community member     | A Status user that is part of a Status community |
+| Community owner node | A Status node with message archive capabilities enabled, run by a community owner |
+| Community member node| A Status node with message archive capabilities enabled, run by a community member |
+| Live messages        | Waku messages received through the Waku network |
+
+
+## Supported platforms
+
+Creating, distributing, downloading and unpacking community history archives SHOULD only be supported in desktop clients. Mobile clients SHOULD NOT implement this functionality due to bandwidth and storage limitations.
+
+## Requirements / Assumptions
+
+This specification has the following assumptions:
+
+- Store nodes are available 24/7, ensuring constant live message availability
+- Store nodes have enough storage to persist historical messages for up to 30 days
+- No store nodes have storage to persist historical messages older than 30 days
+- All nodes are honest
+- The network is reliable
+
+Furthermore, it assumes that:
+
+- Community owner nodes have enough storage to persist historical messages older than 30 days
+- Community owner nodes provide archives with historical messages **at least** every 30 days
+- Community owner nodes receive all community messages
+- Community owner nodes are honest
+
+## Overview
+
+The following is a high-level overview of the user flow and features this specification describes.
+
+### Serving community history archives
+
+Community owner nodes go through the following (high level) process to provide community members with message histories (assumes community owner node is available 24/7):
+
+1. Community owner creates a Status community
+2. Community owner enables community history archive support
+3. A special type of channel for distributing magnet links ([Magnet URI scheme](https://en.wikipedia.org/wiki/Magnet_URI_scheme), [Extensions for Peers to Send Metadata Files](https://www.bittorrent.org/beps/bep_0009.html)) is created
+4. Community owner invites members and creates additional channels
+5. Community owner node receives messages and stores them into local database
+6. After 7 days, the community owner node exports and compresses last 7 days worth of messages from database and creates a magnet link from that data via torrent client
+7. Community owner node creates message archive index and bundles the previously generated magnet link into it
+8. Community owner node creates magnet link from index and distributes it to community members via special channel created in step 2) through the Waku network
+9. Every subsequent 7 days, the step 6) - 8) are repeated and bundled with the previously distributed archive
+
+### Serving archives for missed messages
+
+If the community owner node goes offline, it MUST go through the following process:
+
+1. Community owner node restarts
+2. Community owner node requests messages from store nodes for the missed time range
+3. Missed messages are stored into local database
+4. Community owner node creates message archive and magnet link for missed messages
+5. Community oowner node creates new message archive index bundled with newly created magnet link
+6. The new index is is distributed to community members via special channel through the Waku network
+
+### Receiving community history archives
+
+Community member nodes go through the following (high level) process to fetch and restore community message histories:
+
+1. User joins community and becomes community member
+2. By joining a community, member nodes automatically subscribe to special magnet link channel provided by the community
+3. Member node requests message history (last 30 days) of community channels from store nodes
+4. Member node receives magnet link message from store nodes
+5. Member node extracts magnet link from message and passes it to torrent client
+6. Torrent client downloads latest message archive index via magnet link
+7. Member node fetches missing archives via torrent
+8. Member node unpacks and decompresses message archive data to then hydrate its local database
+
+## Storing live messages
+
+Community owner nodes MUST store live messages as [14/WAKU2-MESSAGE](https://rfc.vac.dev/spec/14/). This is required to provide confidentiality, authenticity, and integrity of message data distributed via the BitTorrent layer, and later validated by Status nodes when they unpack message history archives.
+
+Community owner nodes SHOULD remove those messages from their local databases after they have been turned into archives and distributed to the BitTorrent network.
+
+## Exporting messages for bundling
+
+Community owner nodes export messages from their local database for creating and bundling history archives using the following criteria:
+
+- Messages to be exported MUST have a `contentTopic` that matches any of the topics of the community channels
+- Messages to be exported MUST have a `timestamp` that lies within a given time range
+
+The range for the `timestamp` depends on the context in which the community owner node attempts to create a history archive. This can be one of the following:
+
+1. The community owner node attempts to create an archive periodically for the past seven days (including the current day). In this case, the `timestamp` has to lie within the day the last time an archive was created and the current day.
+2. The community owner node has been offline and attempts to create an archive for all the live messages it has missed since it went offline. In this case, the `timestamp` has to lie within the day the latest message was received and the current day.
+
+Exported messages MUST be restored as [14/WAKU2-MESSAGE](https://rfc.vac.dev/spec/14/) for bundling. Waku messages that have been exported for bundling can now be removed from the community owner node's database (community owner nodes still maintain a database of application messages).
+
+## Message history archives
+
+Message history archives are represented as `WakuMessageArchive` and created from messages exported from the local database. Message history archives are implemented using the following protocol buffer.
+
+### WakuMessageHistoryArchive
+
+The `from` field SHOULD contain a timestamp of the time range's lower bound.
+
+The `to` field SHOULD contain a timestamp of the time range's the higher bound.
+
+The `contentTopic` field MUST contain the same `contentTopic` that the archive's `messages` have.
+
+The `messages` field MUST contain all messages that belong into the archive given its `from`, `to` and `contentTopic` fields.
+
+```
+syntax = "proto3"
+
+message WakuMessageArchive {
+  uint64 from = 1
+  uint64 to = 2
+  string contentTopic = 3
+  repeated WakuMessage messages = 4 // `WakuMessage` is provided by 14/WAKU2-MESSAGE
+}
+```
+
+## Creating magnet links
+
+Once a message archive is created, the community owner node MUST derive a magnet link following the [Magnet URI scheme](https://en.wikipedia.org/wiki/Magnet_URI_scheme) using the underlying BitTorrent protocol client.
+
+The `dn` parameter ("display name") in the resulting magnet link MAY be optional.
+
+The resulting magnet link MUST be bundled into a `WakuMessageArchiveIndex`, which is then later distributed to other Status nodes.
+
+## Bundling history archives into archive indices
+
+Community owner nodes MUST provide message archives for the entire community history. However, each individual archive only contains a subset of the complete history, that is, either data for a time range of seven days, or, a time range in which the node was offline. Therefore, message history archives need to be bundled into a `WakuMessageArchiveIndex`, which later distributed via the Waku network and allows receiving nodes to fetch archives for individual time ranges.
+
+A `WakuMessageArchiveIndex` is a map where the key is the KECCAK-256 hash of the magnet link derived from the previously created archive and the value an instance of `WakuMessageArchiveMetadata`, which is also derived from the previously created message archive.
+
+### WakuMessageArchiveMetadata
+
+The `from` field SHOULD contain a timestamp of the time range's lower bound.
+
+The `to` field SHOULD contain a timestamp of the time range's the higher bound.
+
+The `contentTopic` field MUST contain the same `contentTopic` that the archive's `messages` have.
+
+The `magnet_uri_hash` field MUST contain the KECCAK-256 hash of the magnet link created from the `WakuMessageArchive`.
+
+The `magnet_uri` field MUST contain the magnet link created from the `WakuMessageArchive`.
+
+```
+syntax = "proto3"
+
+message WakuMessageArchiveMetadata {
+  uint64 from = 1
+  uint64 to = 2
+  string contentTopic = 3
+  string magnet_uri_hash = 4
+  string magnet_uri = 5
+}
+
+message WakuMessageArchiveIndex {
+  map<string, WakuMessageArchiveMetadata> archives = 1
+}
+```
+
+The community owner node MUST create a `WakuMessageArchiveIndex` every time it creates a new `WakuMessageArchive`.
+
+For every created `WakuMessageArchive`, there MUST be a `WakuMessageArchiveMetadata` entry in the index map.
+
+The the community owner node MUST derive a magnet link from the newly created `WakuMessageArchiveIndex` so it can be distributed to community member nodes.
+
+## Message archive distribution
+
+Message archives are available via the BitTorrent network as soon as magnet links for them have been created.
+Other community member nodes will download the message archives from the BitTorrent network once they receive a magnet link that contains a message archive index.
+
+The community owner node MUST send magnet links containing message archive indices to a special community channel. The topic of that special channel consists of the community id followed by the string `-archives`:
+
+```
+{community_id}-archives
+```
+
+All messages sent with this topic MUST be instances of `ApplicationMetadataMessage` ([6/PAYLOADS](/specs/6-payloads)) with a `payload` of `CommunityMessageArchiveIndex`.
+
+Only the community owner has permission to send messages with this topic.
+Community members MUST NOT have permission to send messages with this topic.
+However, community member nodes MUST subscribe to this topic to receive message archive indices.
+
+## Canonical message histories
+
+Only community owners are allowed to distribute messages with magnet links via the magnet link channel. Community members MUST NOT be allowed to distribute magnet links. Since the magnet links are created from the community owner node's database (and previously distributed archives), the message history provided by the community owner becomes the canonical message history and single source of truth for the community.
+
+Community member nodes MUST replace messages in their local databases with the messages extracted from archives within the same time range. Messages that didn't receive the community owner node MUST be removed and are no longer part of the message history of interest, even if it already existed in a community member node's database.
+
+
+## Fetching message archives
+
+Generally, fetching message archives is a tree step process:
+
+1. Receive message archive index signal, download index, then determine which message archives to download
+3. Download individual archives
+
+Community member nodes subscribe to the read-only channel that community owner nodes publish message archive indices to. There are two scenarios in which member nodes can receive message archive index messages:
+
+1. The member node receives it via live messages, that is, messages that are relayed by store nodes 
+2. The member node requests messages for a time range of up to 30 days from store nodes (this is the case when a new community member joins a community)
+
+### Downloading message archive indices
+When member nodes receive a message with a `CommunityMessageArchiveIndex` ([6/PAYLOADS](/specs/6-payloads)) from the aforementioned channnel, they MUST extract the `magnet_uri` and pass it to their underlying BitTorrent client so they can fetch the latest message archive index.
+
+Due to the nature of distributed systems, there's no guarantee that a received message is the "last" message. This is especially true when member nodes request historical messages from store nodes. 
+
+Therefore, member nodes MUST wait for 20 seconds after receiving the last `CommunityMessageArchiveIndex` before they start extracting the magnet link to fetch the latest archive index.
+
+### Downloading individual archives
+Once a message archive index is downloaded, community member nodes use a local lookup table to determine which of the listed archives are missing. For this lookup to work, member nodes MUST store the KECCAK-256 hashes of the magnet links for archives they've downloaded.
+
+Given a list of magnet links provided by the index to download individual archives, member nodes can:
+
+1. **Download all archives** - Extract each magnet link in the index and pass them to the underlying BitTorrent client (this is the case for new community member nodes that haven't downloaded any archives yet)
+2. **Download only the latest archive** - Extract only the newest magnet link and pass it to the BitTorrent client (this the case for any member node that already has downloaded all previous history and is now interested in only the latst archive)
+3. **Download specific archives** - Look into `from` and `to` fields of every `WakuMessageArchiveMetadata` and only extract magnet links for archives of a specific time range (can be the case for member nodes that have recently joined the network and are only interested in a subset of the complete history)
+
+
+## Storing historical messages
+
+When message archives are fetched, community member nodes MUST unwrap the resulting `WakuMessage` instances into `ApplicationMetadataMessage` instances and store them in their local database.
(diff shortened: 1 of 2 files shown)


docs/raw/serving-community-history.md
```diff
@@ -53,19 +53,24 @@ The following terminology is used throughout this specification:
...
 | Live messages        | Waku messages received through the Waku network |
+| BitTorrent client    | A program implementing the [BitTorrent](https://bittorrent.org) protocol |
+| Torrent/Torrent file | A file containing metadata about data to be downloaded by BitTorrent clients |
+| Magnet link          | A link encoding the metadata provided by a torrent file ([Magnet URI scheme](https://en.wikipedia.org/wiki/Magnet_URI_scheme)) |
 
 
 ## Supported platforms
 
-Creating, distributing, downloading and unpacking community history archives SHOULD only be supported in desktop clients. Mobile clients SHOULD NOT implement this functionality due to bandwidth and storage limitations.
+Creating, distributing, downloading and unpacking community history archives SHOULD only be supported in Status desktop clients. Status mobile clients SHOULD NOT implement this functionality due to bandwidth and storage limitations.
```
John-44: I would change this to: Status mobile are out of scope for the MVP of this service.  However we may consider bringing this service to Status Mobile in the future.

Thinking behind this comment: in many countries mobile bandwidth is plentiful, very fast, and very cheep, for example in Ukraine it costs about $5 per month for unlimited mobile data, and the speed of this mobile data is normally between 50 and 100 megabits per second (faster than wired ADSL in London!).  In the US mobile data is both expensive and slow, but for folks who are in countries with fast a cheep mobile data, mobile data usage isn't a worry.

Re. storage, in 64gb seems to be becoming the storage baseline these days, with the latest top of the line phones now offering 1TB storage(!!!).  So storage will probably not be a limitation on mobile in the future.

docs/raw/serving-community-history.md
```diff
@@ -84,200 +89,282 @@ Furthermore, it assumes that:
...
 
 ### Serving community history archives
 
 Community owner nodes go through the following (high level) process to provide community members with message histories (assumes community owner node is available 24/7):
 
-1. Community owner creates a Status community
-2. Community owner enables community history archive support
-3. A special type of channel for distributing magnet links ([Magnet URI scheme](https://en.wikipedia.org/wiki/Magnet_URI_scheme), [Extensions for Peers to Send Metadata Files](https://www.bittorrent.org/beps/bep_0009.html)) is created
+1. Community owner creates a Status community (previously known as [org channels](https://github.com/status-im/specs/pull/151))
+2. Community owner enables community history archive support (possibly on by default but can be turned off as well - see [UI feature spec](https://github.com/status-im/feature-specs/pull/36))
```
John-44: Perhaps change to "Community owner doesn't disable community history archive support"

Remove word "possibly", target is on by default

docs/raw/serving-community-history.md
```diff
@@ -84,200 +89,282 @@ Furthermore, it assumes that:
...
+3. A special type of channel to exchange metadata about the archival data is created
```
John-44: Add: Note this channel is not visible in the UI.

docs/raw/serving-community-history.md
```diff
@@ -84,200 +89,282 @@ Furthermore, it assumes that:
...
+6. Member node downloads [message archive index](#message-history-archive-index) file and determines which message archives are not downloaded yet (all or some)
+7. Member node fetches missing message archive data via torrent
+8. Member node unpacks and decompresses message archive data to then hydrate its local database, deleting any messages for that community that the database previously stored in the same time range as covered by the message history archive
 
 ## Storing live messages
 
 Community owner nodes MUST store live messages as [14/WAKU2-MESSAGE](https://rfc.vac.dev/spec/14/). This is required to provide confidentiality, authenticity, and integrity of message data distributed via the BitTorrent layer, and later validated by Status nodes when they unpack message history archives.
 
-Community owner nodes SHOULD remove those messages from their local databases after they have been turned into archives and distributed to the BitTorrent network.
+Community owner nodes SHOULD remove those messages from their local databases once they are older than 30 days and after they have been turned into message archives and distributed to the BitTorrent network.
```
John-44, line 143: Are you sure this is correct?  If a community owner node removes all messages older than 30 days from it's local database, how will the community owner be able to access messages older than 30 days using the Status desktop UI?

docs/raw/serving-community-history.md
```diff
@@ -84,200 +89,282 @@ Furthermore, it assumes that:
...
 
-The `magnet_uri` field MUST contain the magnet link created from the `WakuMessageArchive`.
+A torrent's source folder MUST contain the following two files:
+
+- `data` - Contains all protobuf encoded message history archives concatenated in ascending order
+- `index` - Contains the protobuf encoded message history archive index
+
+Community owner nodes SHOULD store these files in a dedicated folder that is identifiable via the community id.
+
+Community owner nodes MAY remove older torrent files that were generated for previous message history archives, after a new torrent was created.
```
John-44: Shouldn't the removing of the previous (old) torrent file be automatic after the new torrent file is created?  Is there any reason not to do this?

docs/raw/serving-community-history.md
0x-r4bbit: Happy to change that!

docs/raw/serving-community-history.md
0x-r4bbit: Sounds good!

docs/raw/serving-community-history.md
0x-r4bbit: Can add this. Reason I didn't mentioned anything about it being hidden is because this is a spec for the protocol, which should be UI agnostic.

docs/raw/serving-community-history.md
0x-r4bbit: Ah, we had this same discussion in the other PR :D 

See: https://github.com/status-im/specs/pull/162#discussion_r767980234

This is about removing the Waku messages, not the application messages.

docs/raw/serving-community-history.md
0x-r4bbit: That's a good point. The new/latest torrent file will replace the current one, making this paragraph obsolete.

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,402 @@
...
+
+## Message archive distribution
+
+Message archives are available via the BitTorrent network as they are being [seeded by the community owner node](#seeding-message-history-archives).
+Other community member nodes will download the message archives from the BitTorrent network once they receive a magnet link that contains a message archive index.
+
+The community owner node MUST send magnet links containing message archives and the message archive index to a special community channel. The topic of that special channel follows the following format:
+
+```
+/{application-name}/{version-of-the-application}/{content-topic-name}/{encoding}
```
0x-r4bbit, line 315: Still need to update this. What `content-topic-name` should we use here? @staheri14 thoughts?

docs/raw/serving-community-history.md
John-44: ahh yes, thanks for reminding me! Ignore this comment ;-)

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+## Supported platforms
+
+Creating, distributing, downloading and unpacking community history archives SHOULD only be supported in Status desktop clients. Status mobile are out of scope for the MVP of this service. However we may consider bringing this service to Status Mobile in the future.
+
+
+## Requirements / Assumptions
+
+This specification has the following assumptions:
+
+- Store nodes are available 24/7, ensuring constant live message availability
```
staheri14, line 85: You may add another item "The storage time range limit is 30 days."

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+
+These assumptions are less than ideal and will be enhanced in future work. This [discussion](https://forum.vac.dev/t/status-communities-protocol-and-product-point-of-view/114) provides more information on that.
+
+## Overview
+
+The following is a high-level overview of the user flow and features this specification describes. For more detailed descriptions, read the dedicated sections in this specification.
+
+### Serving community history archives
+
+Community owner nodes go through the following (high level) process to provide community members with message histories (assumes community owner node is available 24/7):
```
staheri14, line 106: This part 
> (assumes community owner node is available 24/7):

can be included in the list of assumptions above.

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+
+1. Community owner creates a Status community (previously known as [org channels](https://github.com/status-im/specs/pull/151))
+2. Community owner doesn't disable community history archive support (on by default but can be turned off as well - see [UI feature spec](https://github.com/status-im/feature-specs/pull/36))
```
staheri14, line 109: Do these two terms mean the same thing? if yes, then I suggest to use a consistent term across the spec
"community history archive support" and "message archive capabilities" which is used in the Terminology table

staheri14, line 109:
```suggestion
2. Community owner enables community history archive support (on by default but can be turned off as well - see [UI feature spec](https://github.com/status-im/feature-specs/pull/36))
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+3. A special type of channel to exchange metadata about the archival data is created, this channel should not be visible in the user interface
+4. Community owner invites members and creates additional channels
```
staheri14, line 111:
```suggestion
4. Community owner invites community members and creates additional channels visible in the user interface
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+5. Community owner node receives messages and stores them into local database
```
staheri14, line 112:
```suggestion
5. Community owner node receives messages published in the channels and stores them into a local database
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+6. After 7 days, the community owner node exports and compresses last 7 days worth of messages from database and bundles it together with a [message archive index](#waku-message-archive-index) into a torrent, from which it then creates a magnet link ([Magnet URI scheme](https://en.wikipedia.org/wiki/Magnet_URI_scheme), [Extensions for Peers to Send Metadata Files](https://www.bittorrent.org/beps/bep_0009.html)) 
+7. Community owner sends the magnet link created in step 6) to community members via special channel created in step 3) through the Waku network
```
staheri14, line 114:
```suggestion
7. Community owner sends the magnet link created in step 6 to community members via special channel created in step 3 through the Waku network
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+8. Every subsequent 7 days, the steps 6) and 7) are repeated and the new message archive data is appended to the previously created message archive data
```
staheri14, line 115:
```suggestion
8. Every subsequent 7 days, steps 6 and 7 are repeated and the new message archive data is appended to the previously created message archive data 
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+- All nodes are honest
+- The network is reliable
+
+Furthermore, it assumes that:
+
+- Community owner nodes have enough storage to persist historical messages older than 30 days
+- Community owner nodes provide archives with historical messages **at least** every 30 days
+- Community owner nodes receive all community messages
+- Community owner nodes are honest
+
```
staheri14, line 97:
```suggestion
- The community owner knows at least one store node from which it can query historical messages if need be
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+8. Every subsequent 7 days, the steps 6) and 7) are repeated and the new message archive data is appended to the previously created message archive data 
+
+### Serving archives for missed messages
+
+If the community owner node goes offline (where "offline" means, the owner node's main process is no longer running), it MUST go through the following process:
+
+1. Community owner node restarts
+2. Community owner node requests messages from store nodes for the missed time range for all channels in their community
+3. All missed messages are stored into community owner node's local message database
+4. If 7 or more days have elapsed since the last message history torrent was created, the community owner node will perform step 6) and 7) of [Serving community history archives](#serving-community-history-archives) for every 7 days worth of messages in the missed time range (e.g. if the node was offline for 30 days, it will create 4 message history archives)
```
staheri14, line 124:
```suggestion
4. If 7 or more days have elapsed since the last message history torrent was created, the community owner node will perform steps 6 and 7 of [Serving community history archives](#serving-community-history-archives) for every 7 days worth of messages in the missed time range (e.g. if the node was offline for 30 days, it will create 4 message history archives)
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+
+### Receiving community history archives
+
+Community member nodes go through the following (high level) process to fetch and restore community message histories:
+
+1. User joins community and becomes community member (see [org channels spec](https://github.com/status-im/specs/pull/151))
+2. By joining a community, member nodes automatically subscribe to special channel for message archive metadata exchange provided by the community
+3. Member node requests live message history (last 30 days) of special community channel from store nodes (in addition to last 7 days of message for all other community channels)
```
staheri14, line 132:
```suggestion
3. Member node requests live message history (last 30 days)  of all the community channels including the special channel from store nodes
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+4. Member node receives waku message ([14/WAKU2-MESSAGE](https://rfc.vac.dev/spec/14/)) that contains the metadata magnet link from the special channel
+5. Member node extracts the magnet link from the waku message and passes it to torrent client
+6. Member node downloads [message archive index](#message-history-archive-index) file and determines which message archives are not downloaded yet (all or some)
+7. Member node fetches missing message archive data via torrent
+8. Member node unpacks and decompresses message archive data to then hydrate its local database, deleting any messages for that community that the database previously stored in the same time range as covered by the message history archive
+
+## Storing live messages
+
+Community owner nodes MUST store live messages as [14/WAKU2-MESSAGE](https://rfc.vac.dev/spec/14/). This is required to provide confidentiality, authenticity, and integrity of message data distributed via the BitTorrent layer, and later validated by Status nodes when they unpack message history archives.
```
staheri14, line 141:
```suggestion
Community owner node MUST store live messages as [14/WAKU2-MESSAGE](https://rfc.vac.dev/spec/14/). This is required to provide confidentiality, authenticity, and integrity of message data distributed via the BitTorrent layer, and later validated by community members when they unpack message history archives.
```

staheri14: How many community owners are assumed? one or more? if only one then we shall use the singular form throughout the spec

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+
+Community owner nodes export messages from their local database for creating and bundling history archives using the following criteria:
+
+- Messages to be exported MUST have a `contentTopic` that match any of the topics of the community channels
+- Messages to be exported MUST have a `timestamp` that lies within a time range of 7 days
+
+The `timestamp` is determined by the context in which the community owner node attempts to create a message history archives as described below:
+
+1. The community owner node attempts to create an archive periodically for the past seven days (including the current day). In this case, the `timestamp` has to lie within those 7 days.
+2. The community owner node has been offline (owner node's main process has stopped and needs restart) and attempts to create archives for all the live messages it has missed since it went offline. In this case, the `timestamp` has to lie within the day the latest message was received and the current day.
```
staheri14, line 155: Shouldn't the whole missed time range be broken into 7-day intervals? I think item 1 already covers it and there is no need for the second item. 
I'd suggest adding the explanation of item 2 (how to measure the missed time range) under the second time of this section https://github.com/status-im/specs/blob/c9184b74b5623fe14aa4bd99ed8b268a65984117/docs/raw/serving-community-history.md#serving-archives-for-missed-messages

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+
+## Storing live messages
+
+Community owner nodes MUST store live messages as [14/WAKU2-MESSAGE](https://rfc.vac.dev/spec/14/). This is required to provide confidentiality, authenticity, and integrity of message data distributed via the BitTorrent layer, and later validated by Status nodes when they unpack message history archives.
+
+Community owner nodes SHOULD remove those messages from their local databases once they are older than 30 days and after they have been turned into message archives and distributed to the BitTorrent network.
+
+## Exporting messages for bundling
+
+Community owner nodes export messages from their local database for creating and bundling history archives using the following criteria:
```
staheri14, line 147:
```suggestion
Community owner nodes export Waku messages from their local database for creating and bundling history archives using the following criteria:
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+
+- Messages to be exported MUST have a `contentTopic` that match any of the topics of the community channels
```
staheri14, line 149:
```suggestion
- Waku messages to be exported MUST have a `contentTopic` that match any of the topics of the community channels
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+- Messages to be exported MUST have a `timestamp` that lies within a time range of 7 days
```
staheri14, line 150:
```suggestion
- Waku messages to be exported MUST have a `timestamp` that lies within a time range of 7 days
```

docs/raw/serving-community-history.md
staheri14, line 141:
```suggestion
For the archival data serving, community owner nodes MUST store live messages as [14/WAKU2-MESSAGE](https://rfc.vac.dev/spec/14/). This is in addition to their database of application messages. This is required to provide confidentiality, authenticity, and integrity of message data distributed via the BitTorrent layer, and later validated by Status nodes when they unpack message history archives.
```

staheri14 (commented):
Thank you Pascal for the spec, overall looks good to me!
I have left some comments and suggestions. Did not get to review the second half of the spec, will do it during the week and leave further comments.

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+Exported messages MUST be restored as [14/WAKU2-MESSAGE](https://rfc.vac.dev/spec/14/) for bundling. Waku messages that are older than 30 days and have been exported for bundling can be removed from the community owner node's database (community owner nodes still maintain a database of application messages).
+
+## Message history archives
+
+Message history archives are represented as `WakuMessageArchive` and created from waku messages exported from the local database. Message history archives are implemented using the following protocol buffer.
+
+### WakuMessageHistoryArchive
+
+The `from` field SHOULD contain a timestamp of the time range's lower bound.
+
```
staheri14, line 166:
```suggestion
Its type parallels the `timestamp` field of [`WakuMessage`](https://rfc.vac.dev/spec/14/#payloads).


```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+The `to` field SHOULD contain a timestamp of the time range's the higher bound.
+
```
staheri14, line 168:
```suggestion
Its type parallels the `timestamp` field of [`WakuMessage`](https://rfc.vac.dev/spec/14/#payloads).


```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+The `contentTopic` field MUST contain a list of all communiity channel topics.
+
+The `messages` field MUST contain all messages that belong into the archive given its `from`, `to` and `contentTopic` fields.
+
+The `padding` field MUST contain the amount of zero bytes needed so that the overall byte size of the protobuf encoded `WakuMessageArchive` is a multiple of the `pieceLength` used to divide the message archive data into pieces, as explained in [creating message archive torrents](#creating-message-archive-torrents).
```
staheri14, line 173: "pieces" is unclear at this point, I'd suggest something like below:

The `padding` field MUST contain the amount of zero bytes needed so that the overall byte size of the protobuf encoded `WakuMessageArchive` is a multiple of the `pieceLength`. This is needed for seamless encoding and decoding of archival data in interaction with BitTorrent as explained in [creating message archive torrents](#creating-message-archive-torrents).

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+  uint8 version = 1
+  WakuMessageArchiveMetadata metadata = 2
+  repeated WakuMessage messages = 3 // `WakuMessage` is provided by 14/WAKU2-MESSAGE
+  bytes padding = 4
+}
+```
+
+## Message history archive index
+
+Community owner nodes MUST provide message archives for the entire community history. Each individual archive only contains a subset of the complete history, that is, data for a time range of seven days, and all message history archives are concatenated into a single file as byte string (see [Ensuring reproducible data pieces](#ensuring-reproducible-data-pieces)).
```
staheri14, line 195:
```suggestion
Community owner nodes MUST provide message archives for the entire community history. The entire history consists of a set of `WakuMessageArchive`s where each archive contains a subset of historical`WakuMessage`s for a time range of seven days. All the `WakuMessageArchive`s are concatenated into a single file as a byte string (see [Ensuring reproducible data pieces](#ensuring-reproducible-data-pieces)).
```
```suggestion
Community owner nodes MUST provide message archives for the entire community history. Each individual archive only contains a subset of the complete history, that is, data for a time range of seven days, and all message history archives are concatenated into a single file as byte string (see [Ensuring reproducible data pieces](#ensuring-reproducible-data-pieces)).
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+## Creating message archive torrents
+
+Community owner nodes MUST create a torrent file ("torrent") containing metadata to all message history archives. To create a torrent file, and later serve the message archive data in the BitTorrent network, community owner nodes MUST store the necessary data in dedicated files on the file system.
+
+A torrent's source folder MUST contain the following two files:
+
+- `data` - Contains all protobuf encoded message history archives concatenated in ascending order
+- `index` - Contains the protobuf encoded message history archive index
+
+Community owner nodes SHOULD store these files in a dedicated folder that is identifiable via the community id.
```
staheri14, line 232: Community id is not defined

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+The community owner node MUST update the `WakuMessageArchiveIndex` every time it creates one or more `WakuMessageArchive`s and bundle it into a new torrent (**TODO: see section**).
+For every created `WakuMessageArchive`, there MUST be a `WakuMessageArchiveIndexMetadata` entry in the `archives` field `WakuMessageArchiveIndex`.
+
+## Creating message archive torrents
+
+Community owner nodes MUST create a torrent file ("torrent") containing metadata to all message history archives. To create a torrent file, and later serve the message archive data in the BitTorrent network, community owner nodes MUST store the necessary data in dedicated files on the file system.
+
+A torrent's source folder MUST contain the following two files:
+
+- `data` - Contains all protobuf encoded message history archives concatenated in ascending order
```
staheri14, line 229: Maybe this

```suggestion
- `data` - Contains all protobuf encoded `WakuMessageArchive`s (as bit strings) concatenated in ascending order based on their time
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+- `index` - Contains the protobuf encoded message history archive index
```
staheri14, line 230:
```suggestion
- `index` - Contains the protobuf encoded `WakuMessageArchiveIndex`.
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+
+Community owner nodes SHOULD store these files in a dedicated folder that is identifiable via the community id.
+
+### Ensuring reproducible data pieces
+
+The community owner node MUST ensure that the byte string resulting from the protobuf encoded `data` is equal to the byte string `data` from the previously generated message archive torrent, plus the data of the latest 7 days worth of messages encoded as `WakuMessageArchive`. Therefore, the size of `data` grows every seven days as it's append only.
+
+The community owner nodes also MUST ensure that the byte size of every individual `WakuMessageArchive` encoded protobuf is a multiple of `pieceLength: ???` (**TODO**) using the `padding` field. If the last piece of a message archive has fewer bytes than `pieceLength`, it MUST be filled with zero bytes until it has the size `pieceLength`.
```
staheri14, line 238: As the `piece` is not properly introduced yet, I suggest the following edits:
```suggestion
The community owner nodes also MUST ensure that the byte size of every individual `WakuMessageArchive` encoded protobuf is a multiple of `pieceLength: ???` (**TODO**) using the `padding` field. If the protobuf encoded 'WakuMessageArchive` is not a multiple of `pieceLength`, its `padding` field MUST be filled with zero bytes and the `WakuMessageArchive` MUST be re-encoded until its size becomes multiple of `pieceLength`.
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+
+This is necessary because message history archive data will be split into pieces of `pieceLength` when the torrent file is created, and the SHA1 hash of every piece is then stored in the torrent file and later used by other nodes to request the data for each individual data piece.
```
staheri14, line 240:
```suggestion
This is necessary because the content of the `data` file will be split into pieces of `pieceLength` when the torrent file is created, and the SHA1 hash of every piece is then stored in the torrent file and later used by other nodes to request the data for each individual data piece.
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+20 #3 #3 #3 #3 #3 #3 #3 #3 #3 // piece[2] SHA1: 0xeef
+#3 #3 #3 #3 #3 #3 #3 #3 #3 #3 // piece[3]
+```
+
+By filling up the remainding space of the third piece with A2 using its `padding` field, it is guaranteed that its SHA1 will stay the same:
+
+```
+ 0 11 22 33 44 55 66 77 88 99 // piece[0] SHA1: 0x123
+10 11 12 13 14 15 16 17 18 19 // piece[1] SHA1: 0x456
+20  0  0  0  0  0  0  0  0  0 // piece[2] SHA1: 0x999
```
staheri14, line 292: The `padding` has zeros, but do we also know its protobuf encoding would result zeros?

staheri14 (commented):
Added more suggestions and questions

docs/raw/serving-community-history.md
0x-r4bbit: Actually, this change doesn't necessarily seem to apply. AFAIK messages are generally only fetched for last 7 days for all community channels. We actually need to change this so that the range is expanded to 30 days for the special channel, to ensure community members receive the latest published magnet link.

@richard-ramos Can you confirm the 7 day rule per channel?

docs/raw/serving-community-history.md
0x-r4bbit: Hard to answer. While the spec assumes a single owner, also to have a single source of truth, it's possible for owners to "transfer ownership" of communities to other members.

This is a point I've mentioned under "Considerations"

docs/raw/serving-community-history.md
0x-r4bbit: Okay, will update this. I thought it would be good to be explicit about the scenario that the node can be offline, in which case there can be up to 30 days worth of messages from which the node needs to create archives (4, 1 for each 7 days)

docs/raw/serving-community-history.md
0x-r4bbit: Can you elaborate what it means for it to "paralell" the field?

docs/raw/serving-community-history.md
0x-r4bbit: You mean it should be added to the terminology section?

docs/raw/serving-community-history.md
0x-r4bbit: It does

0x-r4bbit:
Thanks @staheri14 for all the suggestions!!

Will update the PR accordingly. One thing we need to clarify is what `pieceLength` should be. This value is usually calculated based on the overall data size. For this spec to work, we need to decide and agree on a `pieceLength` regardless of the data size. 

Any ideas?

docs/raw/serving-community-history.md
John-44: Of course a Community owner doesn't *have* to create any additional channel, they could have a community that only contains a single channel.  Whenever a community is created, by default after it's created it always contains a single channel

docs/raw/serving-community-history.md
staheri14: This is my understanding, please let me know if the process is different than this. 
- For **Live messages**, there is no need for a magnet link, they can be fetched from store nodes. It was also implied by the following sentence mentioned at the beginning of the specs.
>Store nodes are available 24/7, ensuring constant live message availability
- Once the live messages are fetched, among them there would be some magnet links. The community member then uses the most recent magnet link and proceeds as normal (fetches the 7-day archives and deletes part of its database that corresponds to the time range covered by each archive)

> AFAIK messages are generally only fetched for last 7

Even for 7-days, it might be the case that there is no magnet link published yet, and the node needs to query the store node.

docs/raw/serving-community-history.md
staheri14: I'd suggest also adding it at the beginning of the spec, it remains a question for the reader until the consideration sections. up to you!

docs/raw/serving-community-history.md
staheri14: It has the same type as the `timestamp` field of the waku message.

docs/raw/serving-community-history.md
staheri14: Yes, indeed

staheri14:
> One thing we need to clarify is what pieceLength should be. This value is usually calculated based on the overall data size. For this spec to work, we need to decide and agree on a pieceLength regardless of the data size. Any ideas?

@PascalPrecht 

I imagine a smaller piece size is good for resource-limited members to easily get started and to contribute, but the connection overhead and bandwidth usage would be higher (to keep track of pieces and to announce ownerships). In contrast, a larger piece size contributes to lowering connection and bandwidth overhead, though we should be mindful of the resource limit of members. 
I'll think about what can be a meaningful `pieceLength` (although, on the spec level, we may just recommend a value rather than fixing it.)

There might be studies around this topic as well.

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+
+The community owner nodes also MUST ensure that the byte size of every individual `WakuMessageArchive` encoded protobuf is a multiple of `pieceLength: ???` (**TODO**) using the `padding` field. If the last piece of a message archive has fewer bytes than `pieceLength`, it MUST be filled with zero bytes until it has the size `pieceLength`.
+
+This is necessary because message history archive data will be split into pieces of `pieceLength` when the torrent file is created, and the SHA1 hash of every piece is then stored in the torrent file and later used by other nodes to request the data for each individual data piece.
+
+By fitting message archives into a multiple of `pieceLength` and ensuring they fill possible remainding space with zero bytes, community owner nodes prevent the **next** message archive to occupy that remainding space of the last piece, which will result in a different SHA1 hash for that piece.
```
staheri14, line 242:
```suggestion
By fitting message archives into a multiple of `pieceLength` and ensuring they fill possible remaining space with zero bytes, community owner nodes prevent the **next** message archive to occupy that remaining space of the last piece, which will result in a different SHA1 hash for that piece.
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+
+With a `pieceLength` of 10 bytes, A2 will fit into `21 / 10 = 2` pieces. The remainder will introduce a third piece:
+
+```
+ 0 11 22 33 44 55 66 77 88 99 // piece[0] SHA1: 0x123
+10 11 12 13 14 15 16 17 18 19 // piece[1] SHA1: 0x456
+20                            // piece[2] SHA1: 0x789
+```
+
+The next `WakuMessageArchive` "A3" will be appended ("#3") to the existing data and occupy the remainding space of the third data piece. The piece at index 2 will now produce a different SHA1 hash:
```
staheri14, line 278:
```suggestion
The next `WakuMessageArchive` "A3" will be appended ("#3") to the existing `data` and occupy the remaining space of the third data piece. The piece at index 2 will now produce a different SHA1 hash:
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+
+```
+ 0 11 22 33 44 55 66 77 88 99 // piece[0] SHA1: 0x123
+10 11 12 13 14 15 16 17 18 19 // piece[1] SHA1: 0x456
+20 #3 #3 #3 #3 #3 #3 #3 #3 #3 // piece[2] SHA1: 0xeef
+#3 #3 #3 #3 #3 #3 #3 #3 #3 #3 // piece[3]
+```
+
+By filling up the remainding space of the third piece with A2 using its `padding` field, it is guaranteed that its SHA1 will stay the same:
```
staheri14, line 287:
```suggestion
By filling up the remaining space of the third piece with A2 using its `padding` field, it is guaranteed that its SHA1 will stay the same:
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+ 0 11 22 33 44 55 66 77 88 99 // piece[0] SHA1: 0x123
+10 11 12 13 14 15 16 17 18 19 // piece[1] SHA1: 0x456
+20  0  0  0  0  0  0  0  0  0 // piece[2] SHA1: 0x999
+#3 #3 #3 #3 #3 #3 #3 #3 #3 #3 // piece[3]
+#3 #3 #3 #3 #3 #3 #3 #3 #3 #3 // piece[4]
+```
+
+## Seeding message history archives
+
+The community owner node MUST seed the [generated torrent](#creating-message-archive-torrents) until a new message history archive is created.
```
staheri14, line 299:
```suggestion
The community owner node MUST seed the [generated torrent](#creating-message-archive-torrents) until a new message history archive `WakuMessageArchive` is created.
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+
+message WakuMessageArchiveIndex {
+  map<string, WakuMessageArchiveIndexMetadata> archives = 1
+}
+```
+
+The community owner node MUST update the `WakuMessageArchiveIndex` every time it creates one or more `WakuMessageArchive`s and bundle it into a new torrent (**TODO: see section**).
+For every created `WakuMessageArchive`, there MUST be a `WakuMessageArchiveIndexMetadata` entry in the `archives` field `WakuMessageArchiveIndex`.
+
+## Creating message archive torrents
```
staheri14, line 223: The structure of the torrent file is not explained here, isn't it needed?

docs/raw/serving-community-history.md
staheri14: > dedicated folder that is identifiable via the community id.

Can you please elaborate on this? the folder should have the same name as the community id?

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+
+## Seeding message history archives
+
+The community owner node MUST seed the [generated torrent](#creating-message-archive-torrents) until a new message history archive is created.
+
+The community owner node SHOULD NOT seed torrents for older message history archives. Only one torrent at a time should be seeded.
+
+## Creating magnet links
+
+Once a torrent file for a message archive is created, the community owner node MUST derive a magnet link following the [Magnet URI scheme](https://en.wikipedia.org/wiki/Magnet_URI_scheme) using the underlying BitTorrent protocol client.
```
staheri14, line 305: >Once a torrent file for a message archive is created

for `a message archive` or `all message archives`?

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+Message archives are available via the BitTorrent network as they are being [seeded by the community owner node](#seeding-message-history-archives).
+Other community member nodes will download the message archives from the BitTorrent network once they receive a magnet link that contains a message archive index.
+
+The community owner node MUST send magnet links containing message archives and the message archive index to a special community channel. The topic of that special channel follows the following format:
+
+```
+/{application-name}/{version-of-the-application}/{content-topic-name}/{encoding}
+```
+
+All messages sent with this topic MUST be instances of `ApplicationMetadataMessage` ([6/PAYLOADS](/specs/6-payloads)) with a `payload` of `CommunityMessageArchiveIndex`.
```
staheri14, line 318: `CommunityMessageArchiveIndex` is going to be added? could not find it

staheri14: how about the type of message? I think by `CommunityMessageArchiveIndex` you meant it is the `type` of `ApplicationMetadataMessage`?

staheri14: What happened to the magnet link? shouldn't it be published in the channel?

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+
+## Canonical message histories
+
+Only community owners are allowed to distribute messages with magnet links via the special channel for magnet link exchange. Community members MUST NOT be allowed to post any messages to the special channel.
+
+Status nodes MUST ensure that any message that isn't signed by the community owner in the special channel is ignored.
+
+Since the magnet links are created from the community owner node's database (and previously distributed archives), the message history provided by the community owner becomes the canonical message history and single source of truth for the community.
+
+Community member nodes MUST replace messages in their local databases with the messages extracted from archives within the same time range. Messages that didn't receive the community owner node MUST be removed and are no longer part of the message history of interest, even if it already existed in a community member node's database.
```
staheri14, line 332:
```suggestion
Community member nodes MUST replace messages in their local databases with the messages extracted from archives within the same time range. Messages that the community owner node didn't receive  MUST be removed and are no longer part of the message history of interest, even if it already existed in a community member node's database.
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+
+
+## Fetching message history archives
+
+Generally, fetching message history archives is a tree step process:
```
staheri14, line 337:
```suggestion
Generally, fetching message history archives is a three step process:
```

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+
+1. Receive message archive index magnet link as described in [Message archive distribution], download `index` file from torrent, then determine which message archives to download
```
staheri14, line 339: I'd suggest linking to the "Creating message archive torrents" section for this part:
> download `index` file from torrent

docs/raw/serving-community-history.md
```diff
@@ -0,0 +1,406 @@
...
+3. Download individual archives
+
+Community member nodes subscribe to the special channel that community owner nodes publish magnet links for message history archives to. There are two scenarios in which member nodes can receive such a magnet link message from the special channel:
+
+1. The member node receives it via live messages, that is, messages that are relayed by store nodes
```
staheri14, line 344: Store nodes are known for their storage service as part of which they also relay messages by running waku relay protocol. So, store nodes are a subset of relay nodes but not all of them. Waku relay nodes are the ones that run the relay protocol. 
I think it might be better to skip this explanation as it may cause confusion for the readers.

```suggestion
1. The member node receives it via live messages by listening to the special channel
```

PR #159, opened 2021-09-10
18/URL scheme: draft

saledjenic:

PR #169, opened 2023-02-20
17/URL data: draft

felicio:
Relates to:

- https://github.com/status-im/status-web/issues/327
- https://github.com/status-im/specs/pull/159

### DIFF
--- a/docs/draft/17-url-data.md
+++ b/docs/draft/17-url-data.md
@@ -83,8 +83,6 @@ message User {
 message URLData {
  // Community, Channel, or User
  bytes content = 1;
- // content and public key concatanated and sha256 hashed twice, returning first 4 bytes
- bytes checksum = 2;
 }
 ```
 
--- a/docs/draft/17-url-data.md
+++ b/docs/draft/17-url-data.md
@@ -41,7 +41,7 @@ This document describes serialization, compression and encoding of data within U
 
 ### Product
 
-- Encode data in URL until Waku response is instant and considered reliable enough for this use case, or until a product roadmap for alternative decentrilzed storage is finalized
+- Encode data in URL until Waku response is instant and considered reliable enough for this use case, or until a product roadmap for alternative decentralized storage is finalized
 - Verify state while loading website until encoded data is matched against Waku response to mitigate identity attacks
 - Accept only `[A-Za-z0-9_-.\u0020]` character class for textual fields
 - Shortest result possible
--- a/docs/draft/17-url-data.md
+++ b/docs/draft/17-url-data.md
@@ -64,6 +64,7 @@ message Community {
  string description = 2;
  uint32 members_count = 3;
  string color = 4;
+ repeated uint32 tag_indices = 5;
 }
 
 message Channel {
--- a/docs/draft/17-url-data.md
+++ b/docs/draft/17-url-data.md
@@ -15,7 +15,7 @@ layout: default
 
 ## Abstract
 
-This document describes serialization, compression and encoding of data within URLs, so the app links could carry enough information to help users with instant previewing and verifying of their contents even prior clicking or taking and action on the content.
+This document describes serialization, compression and encoding of data within URLs, so the app links could carry enough information to help users with instant previewing and verifying of their contents even prior clicking or taking an action on the content.
 
 ## Related scope
 
@@ -48,9 +48,9 @@ This document describes serialization, compression and encoding of data within U
 
 ## Implementation
 
-## Example
+### Example
 
-- See <https://github.com/felicio/status-web/blob/34535355d46593179c997ec5ef0eda32547890aa/packages/status-js/src/utils/encode-url-data.test.ts>
+- See <https://github.com/status-im/status-web/pull/345/files>
 
 ### Data
 
@@ -59,25 +59,25 @@ This document describes serialization, compression and encoding of data within U
 ```protobuf
 syntax = "proto3";
 
-message CommunityPreview {
- string display_name = 3;
- string description = 4;
- uint32 members_count = 5;
- string color = 8;
+message Community {
+ string display_name = 1;
+ string description = 2;
+ uint32 members_count = 3;
+ string color = 4;
 }
 
-message ChannelPreview {
- string display_name = 2;
- string description = 3;
- string emoji = 5;
- string color = 6;
- CommunityPreview community = 7;
+message Channel {
+ string display_name = 1;
+ string description = 2;
+ string emoji = 3;
+ string color = 4;
+ Community community = 5;
 }
 
-message UserPreview {
- string display_name = 2;
- string description = 3;
- string color = 7;
+message User {
+ string display_name = 1;
+ string description = 2;
+ string color = 3;
 }
 
 message URLData {
@@ -86,7 +86,6 @@ message URLData {
  // content and public key concatanated and sha256 hashed twice, returning first 4 bytes
  bytes checksum = 2;
 }
-
 ```
 
 ### Encoding
@@ -109,6 +108,10 @@ message URLData {
 ## Discussions
 
 - See <https://github.com/status-im/status-web/issues/327>
+
+## Proof of concept
+
+- See <https://github.com/felicio/status-web/blob/825262c4f07a68501478116c7382862607a5544e/packages/status-js/src/utils/encode-url-data.compare.test.ts#L4>
 
 ## Footnotes
 
--- a/docs/draft/17-url-data.md
+++ b/docs/draft/17-url-data.md
@@ -88,7 +88,7 @@ message URLData {
 }
 
 // Field on CommunityDescription, CommunityChat and ContactCodeAdvertisement
-message URLProps {
+message URLParams {
  string encoded_url_data = 1;
  // Signature of encoded URL data
  string encoded_signature = 2;
--- a/docs/draft/17-url-data.md
+++ b/docs/draft/17-url-data.md
@@ -15,13 +15,13 @@ layout: default
 
 ## Abstract
 
-This document describes serialization, compression and encoding of data within URLs, so the app links could carry enough information to help users with instant previewing and verifying of their contents even prior clicking or taking an action on the content.
+This document describes serialization, compression and encoding of data within URLs, so the app links could carry enough information to help users with instant previewing and verifying even prior clicking or taking an action on the content.
 
 ## Related scope
 
 ### Specs
 
-- URL scheme
+- See <https://github.com/status-im/specs/pull/159> for URL scheme
 
 ### Features
 
@@ -83,6 +83,13 @@ message User {
 message URLData {
  // Community, Channel, or User
  bytes content = 1;
+}
+
+// Field on CommunityDescription, CommunityChat and ContactCodeAdvertisement
+message URLProps {
+ string encoded_url_data = 1;
+ // Signature of encoded URL data
+ string encoded_signature = 2;
 }
 ```
 
--- a/docs/draft/17-url-data.md
+++ b/docs/draft/17-url-data.md
@@ -72,6 +72,7 @@ message Channel {
  string emoji = 3;
  string color = 4;
  Community community = 5;
+ string uuid = 6;
 }
 
 message User {
--- a/docs/draft/17-url-data.md
+++ b/docs/draft/17-url-data.md
@@ -1,0 +1,118 @@
+---
+permalink: /draft/17
+title: 17/URL data
+parent: Draft specs
+layout: default
+---
+
+# 17/URL data
+
+> Version: 0.1
+>
+> Status: Draft
+>
+> Authors: Felicio Mununga <felicio@status.im>
+
+## Abstract
+
+This document describes serialization, compression and encoding of data within URLs, so the app links could carry enough information to help users with instant previewing and verifying of their contents even prior clicking or taking and action on the content.
+
+## Related scope
+
+### Specs
+
+- URL scheme
+
+### Features
+
+- Onboarding website
+- Link preview
+- Link sharing
+- Deep linking
+- Routing and navigation
+
+### App entities
+
+- Community
+- Channel
+- User
+
+## Requirements
+
+### Product
+
+- Encode data in URL until Waku response is instant and considered reliable enough for this use case, or until a product roadmap for alternative decentrilzed storage is finalized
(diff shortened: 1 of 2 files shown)


docs/draft/17-url-data.md
```diff
@@ -0,0 +1,118 @@
...
+
+- Community
+- Channel
+- User
+
+## Requirements
+
+### Product
+
+- Encode data in URL until Waku response is instant and considered reliable enough for this use case, or until a product roadmap for alternative decentrilzed storage is finalized
```
rymnc:
```suggestion
- Encode data in URL until Waku response is instant and considered reliable enough for this use case, or until a product roadmap for alternative decentralized storage is finalized
```

docs/draft/17-url-data.md
```diff
@@ -0,0 +1,118 @@
...
+}
+
+message URLData {
+ // Community, Channel, or User
+ bytes content = 1;
+ // content and public key concatanated and sha256 hashed twice, returning first 4 bytes
+ bytes checksum = 2;
+}
+
+```
```
rymnc: Since this spec is written from scratch ~ would it be possible to store this metadata on ipfs?

```suggestion
#### Fields

```protobuf
syntax = "proto3";

message CommunityPreview {
 bytes community_id = 1;
 bytes ipfs_cid = 2;
}

message ChannelPreview {
 bytes channel_id = 1;
 bytes community_id = 2;
 bytes ipfs_cid = 3;
}

message UserPreview {
 string display_name = 2;
 string description = 3;
 string color = 7;
}

message URLData {
 // Community, Channel, or User
 bytes content = 1;
 // content and public key concatanated and sha256 hashed twice, returning first 4 bytes
 bytes checksum = 2;
}
```

cc: @alrevuelta re dos prevention

this shouldn't be difficult for mobile devices to do, esp. since they don't need waku to find this data.

rymnc (commented):
Wondering if this spec should be created in [vacp2p/rfc](https://github.com/vacp2p/rfc). 
cc: @kaiserd

felicio:
> Wondering if this spec should be created in [vacp2p/rfc](https://github.com/vacp2p/rfc).
> cc: @kaiserd

@rymnc you beat me to raising this question first 🙂, I was just editing another spec. And yeah, please let me know. I wasn't sure what the agreement was and this seemed like an obvious place.

docs/draft/17-url-data.md
felicio: I raised that in https://github.com/status-im/status-web/issues/327#issuecomment-1422764275, however product decided not proceed with IPFS now.

c/o @John-44

John-44: @rymnc  it's not possible to use an IPFS pinning service that's either run by Status or paid for by Status for legal reasons.  Without an IPFS pinning service it would not be sufficiently reliable.

Post MVP we will probably build a decenteralized solution for IPFS pinning that doesn't involve Status/Logos the org in any way, but we can't create a dependency on future infrastructure that doesn't yet exist (and won't exist until post-MVP) for a must-have MVP item (the web preview screens are must-have for MVP)

John-44:
> > Wondering if this spec should be created in [vacp2p/rfc](https://github.com/vacp2p/rfc).
> > cc: @kaiserd
> 
> @rymnc you beat me to raising this question first 🙂, I was just editing another spec. And yeah, please let me know. I wasn't sure what the agreement was and this seemed like an obvious place.

@rymnc the documentation for this URL format should be part of the Status Messaging and Status Communities protocol specs that you are currently working on :-)

felicio:
@John-44 @rymnc to clarify, my question wasn't really about relation/linking, which it is of course somewhat related, but rather about actual placement.

Status currently has two repositories

- https://github.com/status-im/specs
- https://github.com/status-im/feature-specs

which have several WIP, but stable too, specs.

While I couldn't be more appreciative of and thankful that Vac is shaking things up and helping tremendously, I'd suggest

- merging the two above mentioned repos into one
- placing there what's really Status specific
- and making sure all client teams are on board (e.g. w/ an announcement), write up, discuss and maintain these

That it is what I wanted to clear out prior linking or making it part of something else.

John-44:
> 

@felicio https://github.com/status-im/specs is the correct repo for this

felicio:
> @felicio https://github.com/status-im/specs is the correct repo for this

@John-44 then,

1. Should https://github.com/vacp2p/rfc/pull/567 be placed in this repo too?
2. Would you agree with merging https://github.com/status-im/specs (implementation) and https://github.com/status-im/feature-specs (features) into one?

Cc @rymnc

John-44:
> > @felicio https://github.com/status-im/specs is the correct repo for this
> 
> @John-44 then,
> 
> 1. Should [feat(56/STATUS-COMMUNITIES): initial draft vacp2p/rfc#567](https://github.com/vacp2p/rfc/pull/567) be placed in this repo too?
> 2. Would you agree with merging https://github.com/status-im/specs (implementation) and https://github.com/status-im/feature-specs (features) into one?
> 
> Cc @rymnc

@rymnc as you are working on writing the missing Status Communities protocol specs and getting all the Status Messaging protocol specs up to date, are you thinking about either:

A) Planning on moving the legacy Status Messaging protocol specs (that are currently located in the https://github.com/status-im/specs repo) to a new repo as you update them, so that when you are finished the legacy https://github.com/status-im/specs repo can be deleted? 

OR

B) Adding the new Status Communities protocol specs to https://github.com/status-im/specs and updating the Status Messaging protocol specs in place in this repo?

The reason I'm asking is that the Status Communities protocol specs and the Status Messaging protocol specs should be in the same place, and should not be duplicated in two different places.

I have no opinion on (and don't mind) what this single place is, as long as both the Status Communities protocol specs and the Status Messaging protocol specs eventually end up in the same place, and as long as there isn't an old out of date duplicate of the Status Messaging protocol specs hanging around anywhere.

e.g. perhaps we should use the vac rfc repo for everything, and get rid of the https://github.com/status-im/specs repo, or we could also do the other way around (use https://github.com/status-im/specs for everything)

oskarth:
Hey everyone! Some historical context that might be useful:

`status-im/specs` (this repo) was our first effort to specify the Status protocol specs. With `vacp2p/rfc` the initial idea was to keep more generic Waku related specs in Vac RFCs, and then have Status protocol specific specs in this repo owned by the Status app teams. Since then, unfortunately keeping Status protocol specs became less of a priority for the app teams. At the same time, feature specs was seen as something that is more useful for Status app team development work, hence `status-im/feature-specs`. These feature specs were created by the Desktop team, and my understanding is that they do not attempt to be protocol specifications, but more about end users and to help developers build an app, as opposed to a protocol.

Since then, the need for protocol specifications has again become clear, especially when it comes to Status Communities and Messaging specs in terms of understanding its scalability, privacy etc properties. Based on this, @rymnc took over the work to specify these into Vac RFCs as "application layer specs" (they are separate from Waku base layer specs), e.g. see  There have also been similar application layer specs that have been defined as Vac RFCs (tagged as follows, https://github.com/vacp2p/rfc/pull/567/files#diff-bd54c3ccca130b9f7dbd9fd6855c36e4feecae4d85ccc09709bda9fec85dd57fR7).

Based on this, I think it is safe to say that this repo is largely deprecated. With Vac RFCs we have a clear workflow for specs, and there are lot of people there who are used to working with protocol specs, so it makes more sense to keep them there. Note that this doesn't mean Status people can't be editors of these specs, and we'd encourage Status people to be editors of specs since you guys are the domain experts.

For the time being though, this is @rymnc and he is working on porting and updating the core Status messaging protocol specs from this repo (including Community spec draft) to Vac RFCs. Once this is done, it makes sense to put a notice/remove those specs in this repo to indicate that this repo isn't kept up to date.

If we consider things like URL data part of the core protocol, then this should indeed be part of the Community/Messaging protocol or one of its companion specs. If it is more of a client specific thing and it doesn't significantly impact core semantics, I think it is OK to just specify it at a user/app level as is done with feature-specs. I'd expect the feature-specs to refer to the protocol specs in terms of core semantics, and in some cases there might be some overlap with a different focus (protocol view or app/end-user view).

As for the other specs that are currently in this repo, if we consider them core protocol specs I suggest we move them to Vac RFCs too (repo owned by Vac, but anyone can be a editor/contributor), and if they are more app/user specific (some of the older specs are indeed of this form) they can be moved to feature-specs (repo by Status).

Also cc @kaiserd

felicio:
Thank you for the much needed context. I will have this spec reviewed and if approved moved.

The following is just me expanding on what was my point of view:

> not attempt to be protocol specifications, but more about end users and to help developers build an app

I looked at the latter as driving the former. The "what" and "why" to the "how". And the closer together in placement (i.e. one repo) and processes (i.e. format, wider reaching team reviews) the better for recalling, reasoning, presenting and maintaining.

> took over the work to specify these

I didn't know all or most of these were to be ported.

> to put a notice/remove those specs in this repo to indicate that this repo isn't kept up to date.

I thought marking those as deprecated/unmaintained/outdated first, announcing it to the teams and revising them in place would help with awareness, prevent scattering of the information and actually cultivate the habit.

docs/draft/17-url-data.md
```diff
@@ -0,0 +1,121 @@
...
+ string display_name = 1;
+ string description = 2;
+ string color = 3;
+}
+
+message URLData {
+ // Community, Channel, or User
+ bytes content = 1;
+ // content and public key concatanated and sha256 hashed twice, returning first 4 bytes
+ bytes checksum = 2;
```
felicio: Not worth it, in my opinion. I expanded on this at https://github.com/status-im/status-web/pull/345#discussion_r1113140636.

> This demonstrates one way of preventing
> 
> - you creating a new community/profile with offending content (e.g. name, description)
> - copying the generated share link
> - slapping my/mine public key at the end
> - and pointing others to the linka and telling them I'm the author of the content
>
> However, if you knew my public key before hand and spent a little effort by generating the link programmatically, this wouldn't catch it. And that is also the reason why I think this checksum is NOT even worth – can't assume the data is valid.

felicio: @cammellos confirmed, replacing with a signature after `#` and adding corresponding field to `CommunityDescription` and `ContactCodeAdvertisement`.

docs/draft/17-url-data.md
```diff
@@ -0,0 +1,126 @@
...
+syntax = "proto3";
+
+message Community {
+ string display_name = 1;
+ string description = 2;
+ uint32 members_count = 3;
+ string color = 4;
+}
+
+message Channel {
```
felicio, line 70: Should we add uuid here, instead of it being a path segment separate from the encoded data and its signature?

docs/draft/17-url-data.md
```diff
@@ -0,0 +1,126 @@
...
+message URLData {
+ // Community, Channel, or User
+ bytes content = 1;
+}
+
+// Field on CommunityDescription, CommunityChat and ContactCodeAdvertisement
+message URLProps {
+ string encoded_url_data = 1;
+ // Signature of encoded URL data
+ string encoded_signature = 2;
```
felicio, line 94: ℹ️ A signature of the already encoded data so the verification does not require decoding, decompressing and deserializing.

docs/draft/17-url-data.md
```diff
@@ -0,0 +1,126 @@
...
+}
+
+message URLData {
+ // Community, Channel, or User
+ bytes content = 1;
+}
+
+// Field on CommunityDescription, CommunityChat and ContactCodeAdvertisement
+message URLProps {
+ string encoded_url_data = 1;
```
felicio, line 92: ℹ️ Stores the url data already encoded alongside the other proto messages so clients don't have to do the additional computation for other members/communities, just mapping now.

docs/draft/17-url-data.md
cammellos: The uuid of the channel?

felicio: For example `30804ea7-bd66-4d5d-91eb-b2dcfe2515b3`. Channel ids are concatenations of uuids and community's public keys.

felicio: Here https://github.com/status-im/specs/pull/159#discussion_r1111762125 are the current proposed changes to the scheme spec with the uuid still in the url in plain text.

cammellos: if you keep it outside the signature, it can be controlled potentially by an attacker, not clear what the benefit would be if everything else is there, but that's the kind of things that you want it to be in the signature just to be safe

felicio: That's why I raised it, thanks. Updated the spec too.

cammellos (approved):

docs/draft/17-url-data.md
felicio: @corpetty @0xc1c4da to ensure integrity of the links containing encoded data at all times we've replaced the checksum for signature.

Is it okay?

Signature was originally considered at https://github.com/status-im/status-web/issues/327#issuecomment-1422764275.

Cc @John-44

docs/draft/17-url-data.md
```diff
@@ -0,0 +1,127 @@
...
+- See <https://github.com/status-im/status-web/pull/345/files>
+
+### Data
+
+#### Fields
+
+```protobuf
+syntax = "proto3";
+
+message Community {
```
felicio, line 62: It's been requested to add tags:

> Add tags to the URL (the maximum number of tags is four and it's just a number from 0 to 70*)

felicio: @John-44 @prichodko addressed in

- https://github.com/status-im/specs/pull/169/commits/44ad67ad5064496cd2678025d3cc27459cf9934c
- https://github.com/status-im/status-web/pull/345/commits/35af4bcf2e1114feb9255aa89f1d6ec1dfbda0f0

However, afaik, there's only 52 tags so I don't know where the 70 comes from. So please, let me know if some are missing so the following lists can be updated too.

- https://github.com/status-im/status-go/blob/b8fd999/protocol/requests/community_tags.go
- https://github.com/status-im/status-web/blob/058d989/packages/status-js/src/preview-client/tags.ts

John-44: If 52 tags have been counted then that is the accurate number (70 tags was just a rough guess, if they have been counted then the counted number is the correct number).

felicio:
Moved to https://github.com/vacp2p/rfc/pull/600.

